PR #8, opened 2016-07-21
use numpy.issubdtype instead of equality for numericness

bollwyvl:
Apparently, windows `int` and `float` are not `==` to most dtypes, causing two test fails:
https://ci.appveyor.com/project/conda-forge/staged-recipes/build/1.0.4847/job/iqs0wu1u7jcamxs3#L418

This adds use of `numpy.issubdtype` for an (apparently) more robust comparison.


wdm0006:
That's a new one for me, but makes some sense, thanks.

PR #20, opened 2017-03-15
Enable a full class transform for OneHotEncoder

justkunz:
@wdm0006 PR to address https://github.com/scikit-learn-contrib/categorical-encoding/issues/19

Have `OneHotEncoder` expand all the classes/categories that were `fit`, instead of just the classes that were passed in to `transform`.

### DIFF
--- a/category_encoders/one_hot.py
+++ b/category_encoders/one_hot.py
@@ -178,8 +178,7 @@ class OneHotEncoder(BaseEstimator, TransformerMixin):
         else:
             return X.values
 
-    @staticmethod
-    def get_dummies(X_in, cols=None):
+    def get_dummies(self, X_in, cols=None):
         """
         """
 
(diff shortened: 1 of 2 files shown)


category_encoders/one_hot.py
```diff
@@ -193,8 +193,10 @@ def get_dummies(X_in, cols=None):
 
         bin_cols = []
         for col in cols:
-            classes = [x for x in set(X[col].values.tolist()) if np.isfinite(x)]
-            for class_ in classes:
+            col_tuples = [class_map['mapping'] for class_map in ordinal_encoder.mapping if class_map['col'] == col][0]
+            fit_classes = set(col_val[1] for col_val in col_tuples)
+            input_classes = set(x for x in set(X[col].values.tolist()) if np.isfinite(x))
+            for class_ in (fit_classes | input_classes):
```
justkunz: If a new class is passed in to transform and handle_unknown == 'impute' then the -1 class is automatically added when `self.ordinal_encoder.transform(X)` is called. That is why I take the union of fit_classes and input_classes (input_classes includes the -1 class, in this specific case).

Otherwise, I could change it to:
```
fit_classes = set(col_val[1] for col_val in col_tuples)
if (self.handle_unknown == 'impute') & (-1 in X[col].unique()):
  fit_classes.update({-1})
for class_ in fit_classes:
```
That way the -1 class is only added when it's present. Either way, a new class column is only added if everything passes correctly through the ordinal encoder.

wdm0006: Ah, sorry for the mixup: I meant to say that we should have a column for the -1 class in the case that handle_unknown is true and there is _not_ an unknown value present, so that the same columns are output from transform no matter what.  So if you are imputing unknown classes with -1, you should _always_ see an output column for that class, just like with any other category. Make sense?

justkunz: Yes, that makes sense. Otherwise there will be the same issues with different sizes in the sklearn pipeline. For a single encoder the output shape should always have the same number of columns, no matter the input. I'll add the change.

wdm0006:
Looks great, Thanks!

PR #24, opened 2017-06-13
Fit with mapping

### LINKED ISSUE #23: mapping with ordinal breaks if fit is called again
this makes it impossible to use in a pipeline where some things need to be fit and ordinal does not iff it has a mapping

This is demonstrated in the test case failing https://github.com/novilabs/categorical-encoding/commit/ddf21f79d17d4aa1f0ba5efd9d516b8e3200f5e1

a86c6f7964:
fixes #23


wdm0006:
Thanks for the PR, this looks good to me.

PR #25, opened 2017-06-13
consistent ordering for categories

a86c6f7964:
it seems better to use the pandas ordering of `Uniques are returned in order of appearance` https://pandas.pydata.org/pandas-docs/stable/generated/pandas.unique.html#pandas.unique rather than something random. This way the models are reproducible

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -219,8 +219,7 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
                     X[switch.get('col')] = X[switch.get('col')].astype(float).reshape(-1, )
         else:
             for col in cols:
-                categories = list(set(X[col].values))
-                random.shuffle(categories)
+                categories = pd.unique(X[col].values)
 
                 X[str(col) + '_tmp'] = np.nan
                 for idx, val in enumerate(categories):


wdm0006:
I think that you're right in that the order should be consistent, for sure. I'm wondering though if order of appearance may be less desirable than something like `sorted(categories)`.  What do you think?

a86c6f7964:
I was thinking that order of appearance would be easier to explain, as opposed to saying that the categorical has to be sortable. I was thinking about trying to do something that was most likely to mutate the data the least when nothing was configured. Maybe there could be a param that allows sorting to be by 'order of appearance' 'sorted' 'order or frequency' with it defaulting to 'order of appearance'?

wdm0006:
Hm, yeah I just have some concern that the order of rows would effect the outcome, which may be reasonable, but ideally, if we shuffled rows we'd get the same thing out regardless.  So I guess your point of making it configurable makes the most sense I think.

a86c6f7964:
My concern is mostly that same data same row order has same predictions. I
think a lot of stuff will change the predictions if the row order changes
(for instance things in say ExtraTrees that are random with a seed)

wdm0006:
Yeah, that makes sense.

a86c6f7964:
is there anything immediate that you would like me to do or does this pull request make sense as is?

wdm0006:
Sorry for the delay, yeah I think this is ok for now, but I would still like to make this choice configurable by the user at some later point. Thank you for the contribution.

PR #29, opened 2017-06-21
Add leave-one-out encoding

hbghhy:
I try to add leave-one-out encoding according to the kaggle and owen-zhang'code. The Randomized part of them is a little confused, so i just add a normal (Gaussian) distribution randomized into it.
@wdm0006


hbghhy:
I used to think user may want to give different row different sigma, so I add it there. But maybe user can do it by set randomized be False, then do it himself. And i just fix it.
@wdm0006

wdm0006:
This looks great thank you, we will need to do some work on the docs and cut a new release soon for this and some of the other recent changes/bugfixes.

Issue #40, opened 2017-11-08
target encoding


rspadim:
should be nice this one too:

https://www.kaggle.com/ogrellier/python-target-encoding-for-categorical-features

wdm0006:
This does look interesting, would you like to help work on a PR for it?

chappers:
There is a link to the actual paper here: https://kaggle2.blob.core.windows.net/forum-message-attachments/225952/7441/high%20cardinality%20categoricals.pdf

intern1knowdis:
@chappers the link does not work :disappointed:

janmotl:
The article is still available from:
	https://dl.acm.org/citation.cfm?id=507538

PR #45, opened 2017-12-20
Add Target Encoder #40

chappers:
As per #40 
Implements Target encoder. Code is based on the leave one out and implements approach described in 

* https://www.slideshare.net/0xdata/feature-engineering-83511751
* https://www.kaggle.com/ogrellier/python-target-encoding-for-categorical-features
* https://kaggle2.blob.core.windows.net/forum-message-attachments/225952/7441/high%20cardinality%20categoricals.pdf

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Encoding Methods
  * Hashing [1]
  * BaseN [6]
  * LeaveOneOut [4]
+ * Target Encoding [7]
 
 Usage
 -----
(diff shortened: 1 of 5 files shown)


wdm0006:
Wow, this looks great, thanks. Could you please add it onto the README and add a reference there as well? Also if possible an entry into the docs under docs/source, you can use another one there as an example.

chappers:
@wdm0006 Done! :)

PR #50, opened 2018-01-08
Wenwu tao

wenwu313:
Hello，I  add inverse_transform in oridinal.py , one_hot.py, binary.py.  if there is any problem, please tell me.
Now I try to add inverse_transform in basen.py


category_encoders/basen.py
```diff
@@ -178,6 +178,53 @@ def transform(self, X, override_return_df=False):
         else:
             return X.values
 
+    def inverse_transform(self, X_in):
```
wdm0006: Should we use Xt as the input name to be more in line with sklearn? http://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html

wenwu313:
Thank you for your advice, I have changed the input name, Because of issue [#52](https://github.com/scikit-learn-contrib/categorical-encoding/issues/52) and [#55](https://github.com/scikit-learn-contrib/categorical-encoding/issues/55), The test code for inverse_transform is not easy to add. I  have tried to add some test code, I think you can merge first, after these issue solved, I will pull request again to modify and add more  test code for inverse_transform.

image
The "None" will transform as "-1", the inverse_transform always raise ValueError

wdm0006:
Ok, that makes sense to me. I'm going to do some issue pruning and will open issues explicitly for those points.

PR #65, opened 2018-05-02
added option to include category values in encoded column names

adamhajari:
@wdm0006 
addresses this issue: https://github.com/scikit-learn-contrib/categorical-encoding/issues/42


wdm0006:
Very cool, I think this will be useful for users. Thanks for the contribution.

Issue #70, opened 2018-05-17
Request to add support for pandas categorical types when using OrdinalEncoder


dennisobrien:
Hi All,

I ran into an issue using OrdinalEncoder where I assumed that it would use the mapping in the categorical types in the passed pandas dataframe.  Instead, it created a mapping based on the order of the data passed to `fit_transform`.  I realize that this is user error, but it would be nice for OrdinalEncoder to use the existing mapping information from the categorical dtype.  If the maintainers think this would be a nice addition, I'd be happy to make a PR.

Here's a demonstration of the issue.

First I'll create a categorical dtype and a dataframe that uses it.
```python
from category_encoders.ordinal import OrdinalEncoder
import pandas as pd
from pandas.api.types import CategoricalDtype

platforms = ['android', 'ios', 'amazon']
platform_category = CategoricalDtype(categories=platforms, ordered=False)

df = pd.DataFrame([
    {'id': 1, 'platform': 'android'},
    {'id': 2, 'platform': 'ios'},
    {'id': 3, 'platform': 'amazon'},
])
df['platform'] = df['platform'].astype(platform_category)
print(df)
```
```
   id platform
0   1  android
1   2      ios
2   3   amazon
```

The encoding from the categorical dtype looks like this:
```python
[(cat, code) for code, cat in enumerate(df['platform'].cat.categories)]
```
```
[('android', 0), ('ios', 1), ('amazon', 2)]
```

Now I'll make an encoder without a mapping parameter, then transform with the data sorted.
```python
cat_encoder_a = OrdinalEncoder(cols=categorical_columns)
df_a = cat_encoder_a.fit_transform(df.sort_values(by='platform', ascending=True))
print(df_a)
```
```
   id  platform
0   1         0
1   2         1
2   3         2
```

The category mapping from the encoder happens to match the categorical dtype mapping because of the sort order of the dataframe.
```python
cat_encoder_a.category_mapping
```
```
[{'col': 'platform', 'mapping': [('android', 0), ('ios', 1), ('amazon', 2)]}]
```

But if I reverse the order of the data passed to `fit_transform` I will get a different mapping.
```python
cat_encoder_b = OrdinalEncoder(cols=categorical_columns)
df_b = cat_encoder_b.fit_transform(df.sort_values(by='platform', ascending=False))
cat_encoder_b.category_mapping
```
```
[{'col': 'platform', 'mapping': [('amazon', 0), ('ios', 1), ('android', 2)]}]
[...]

PR #83, opened 2018-06-17
Add support for pandas categorical types when using OrdinalEncoder

dennisobrien:
Addresses issue #70 

If a column is a pandas categorical dtype and no custom mapping is provided, use the mapping from the pandas categorical.

Added a test that passes when run locally.

Some specific questions I'd appreciate guidance on:
- I added a method `get_mapping_for_col` to utils.py. I'm not sure that the names I used are in line with the project.
- Pandas categoricals will usually fail if impute_missing=True is passed (the default) since it tries a fillna(0) which doesn't always make sense for a pandas categorical.  I didn't touch this, but maybe it deserves special handling.
- Any implementation suggestions?

thanks,
Dennis

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -266,8 +266,12 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
         else:
             mapping_out = []
             for col in cols:
-                categories = [x for x in pd.unique(X[col].values) if x is not None]
-                categories_dict = {x: i + 1 for i, x in enumerate(categories)}
+                if X[col].dtype.name == 'category':
+                    categories = X[col].cat.categories
+                    categories_dict = {x: i for i, x in enumerate(categories)}
+                else:
+                    categories = [x for x in pd.unique(X[col].values) if x is not None]
+                    categories_dict = {x: i + 1 for i, x in enumerate(categories)}
                 X[str(col) + '_tmp'] = X[col].map(lambda x: categories_dict.get(x))
                 del X[col]
                 X.rename(columns={str(col) + '_tmp': col}, inplace=True)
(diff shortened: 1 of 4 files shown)


category_encoders/ordinal.py
```diff
@@ -266,8 +266,12 @@ def ordinal_encoding(X_in, mapping=None, cols=None, impute_missing=True, handle_
         else:
             mapping_out = []
             for col in cols:
-                categories = [x for x in pd.unique(X[col].values) if x is not None]
-                categories_dict = {x: i + 1 for i, x in enumerate(categories)}
+                if X[col].dtype.name == 'category':
+                    categories = X[col].cat.categories
+                    categories_dict = {x: i for i, x in enumerate(categories)}
```
wdm0006: In the normal case we start the categories at 1, not 0 (so, i+1), to reserve 0 for unknown. We used to use -1 for unknown, but some upstream encoders don't support negative categories, so we changed it recently.

category_encoders/ordinal.py
```diff
@@ -281,6 +285,6 @@ def ordinal_encoding(X_in, mapping=None, cols=None, impute_missing=True, handle_
                 except ValueError as e:
                     X[col] = X[col].astype(float).values.reshape(-1, )
 
-                mapping_out.append({'col': col, 'mapping': [(x[1], x[0] + 1) for x in list(enumerate(categories))]}, )
+                mapping_out.append({'col': col, 'mapping': [(cat, code) for cat, code in categories_dict.items()]})
```
wdm0006: for this, mapping can just be `list(categories_dict.items())`

janmotl:
@dennisobrien Support for pandas categorical dtype was implemented. Can you check whether we missed something?

PR #85, opened 2018-06-29
ENH: Performance improvement for BinaryEncoder

### LINKED ISSUE #80: Performance Improvement for Binary Encoder
From issue #74 

@rafarui found that binary encoder could be dramatically sped up by precomputing the encoding on the unique categories then doing a join back to the original data. 

```
uniques = pd.DataFrame(df['col'].unique(), columns = ['col'])
enc = ce.BinaryEncoder(verbose=1, cols=['col'])
uniques = pd.concat([uniques,enc.fit_transform(uniques)], axis=1)
df= df.merge(df, how = 'left', on = 'col')
```

thomasjpfan:
Fixes #80 

For a small test case, I ran the following:

```python
import numpy as np
import pandas as pd
df = pd.DataFrame({'a': np.random.randint(0,high=10, size=10000)})
be = ce.BinaryEncoder(cols=['a'])
```

And timed `be.fit_transform(df)`. On master, it ran in 101 ms, with this PR, it ran in 18ms.

### DIFF
--- a/category_encoders/binary.py
+++ b/category_encoders/binary.py
@@ -118,6 +118,7 @@ class BinaryEncoder(BaseEstimator, TransformerMixin):
             impute_missing=self.impute_missing,
             handle_unknown=self.handle_unknown
         )
+        X = X.drop_duplicates(subset=self.cols) if self.cols else X
         self.ordinal_encoder = self.ordinal_encoder.fit(X)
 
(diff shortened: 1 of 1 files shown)


wdm0006:
Awesome! Thanks, @thomasjpfan

PR #90, opened 2018-07-22
Correct the docstring

LiutongZhou:
If `handle_unknown='impute'`, unknown categories will be encoded to 0 when `transform` is called. 
The docstring says unknown categories will be encoded to -1,  which is wrong.

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -37,7 +37,7 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
     impute_missing: bool
         boolean for whether or not to apply the logic for handle_unknown, will be deprecated in the future.
     handle_unknown: str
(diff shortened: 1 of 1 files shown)


category_encoders/ordinal.py
```diff
@@ -37,7 +37,7 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
     impute_missing: bool
         boolean for whether or not to apply the logic for handle_unknown, will be deprecated in the future.
     handle_unknown: str
-        options are 'error', 'ignore' and 'impute', defaults to 'impute', which will impute the category -1
```
LiutongZhou: -1 should be corrected to be 0 according to the code

wdm0006:
Good catch, thanks

PR #93, opened 2018-08-02
Fix patsy naming issue

adithyabsk:
Columns named with awkward names including restricted characters such as "-" (dashes) will get evaluated to minus signs through patsy which requires their wrapping using the Q builtin.
http://patsy.readthedocs.io/en/latest/builtins-reference.html#patsy.builtins.Q

Minimum broken example:
```python
import pandas as pd
X = pd.DataFrame([1, 2, 3, 4], columns=['test-broken'])
X_trans = ce.HelmertEncoder(cols=['test-broken']).fit_transform(X)
```
Raises the error:
> patsy.PatsyError: Error evaluating factor: NameError: name 'col_test' is not defined
    C(col_test-broken, Helmert)

### DIFF
--- a/category_encoders/backward_difference.py
+++ b/category_encoders/backward_difference.py
@@ -203,7 +203,7 @@ class BackwardDifferenceEncoder(BaseEstimator, TransformerMixin):
 
         bin_cols = []
         for col in cols:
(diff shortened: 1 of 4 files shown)


wdm0006:
Oh great catch, we should add some test cases for this as well. Thanks for the PR

PR #94, opened 2018-08-03
Update target_encoder.py

gsheni:
Changed TargetEncode or TargetEncoder, the proper name.


wdm0006:
doh, good catch, thank you

Issue #107, opened 2018-08-18
Typo in the documentation of TargetEncoder


janmotl:
Typo: TargetEncode -> TargetEncoder

wdm0006:
Hey @janmotl, I've invited you to the org as a maintainer, please continue to make PRs for any significant features (typos are fine to commit directly). I'm traveling quite a bit right now, but will check in here when I can.

janmotl:
Solved with #110

Issue #104, opened 2018-08-16
Encoders that utilize the target should overwrite the default fit_transform()


janmotl:
The default implementation of `fit_transform()` in TransformerMixin looks like:
``` Python
return self.fit(X, y, **fit_params).transform(X)
```

However, if the encoder utilizes the target, we have to use something like:
``` Python
return self.fit(X, y, **fit_params).transform(X, y)
```

If we do not pass the target `y` to `transform()`, the model will unnecessarily overfit.

A quick fix is to add:
``` Python
def fit_transform(self, X, y=None, **fit_params):
    return self.fit(X, y, **fit_params).transform(X, y)
```
into `LeaveOneOutEncoder` and `TargetEncoder`.

janmotl:
Solved with #110

Issue #103, opened 2018-08-09
Move check_random_state in LeaveOneOutEncoder


janmotl:
Following code in LeaveOneOutEncoder:
```
self.random_state_ = check_random_state(self.random_state)
```
should possibly be moved from `fit()` method into `transform()` method, where the random number generator is actually used. Reasoning: I would  expect the following code to pass:
```
enc = encoders.LeaveOneOutEncoder(random_state=2001, randomized=True)
enc.fit(X, y)
a = enc.transform(X_t, y_t)
b = enc.transform(X_t, y_t)
self.assertTrue(a.equals(b))
```
but currently it does not pass the test.

A possible place where to move it:
```
if self.randomized and y is not None:
    random_state_generator = check_random_state(self.random_state)
    X[switch.get('col')] = (X[switch.get('col')] * random_state_generator.normal(1., self.sigma, X[switch.get('col')].shape[0]))
```

wdm0006:
Agreed

janmotl:
Solved with #110

Issue #101, opened 2018-08-09
drop_invariant on string columns


janmotl:
The following code will fail if one of the columns is non-numeric (e.g.: because only some of the string columns were converted):
```
# drop all output columns with 0 variance.
        if self.drop_invariant:
            self.drop_cols = []
            X_temp = self.transform(X)
            self.drop_cols = [x for x in X_temp.columns.values if X_temp[x].var() <= 10e-5]
```

I see multiple options:
1. Measure variance of only numeric columns
2. Measure variance of only generated columns 
3. Measure the length of the `unique()` result
4. Write a function, which compares `X[attribute, i] == X[attribute, i-1]` and returns on the first difference

Discussion:
1. It is necessary to decide whether we want to touch columns that are not transformed (either because they are not in the `col` parameter or because of their data type). I am inclined to not touch them. Reasoning: It would make unit testing with `self.verify_inverse_transform(X_t, enc.inverse_transform(enc.transform(X_t)))` easier. Either way, the documentation should mention whether drop_invariant applies only on generated columns or all columns (even if they contain monstrous or arguably uncomparable objects).
2. Calculation of variance just to learn whether the content of a column is invariant is possibly an overkill. I am not sure whether a loop with a possibility of an early return or a vectorized comparison `col[1:]==col[:-1]` plus handling of the edge cases would be faster.

janmotl:
Solved with #110

PR #110, opened 2018-08-28
Cumulative

janmotl:
This PR fixes some of the reported issues (#107, #104, #103, #101, #95). Furthermore, the documentation now passes doctest (it was necessary to remove the single white line between the example code and the example result,...). And it contains the previous benchmarking code (#46) and WoE (#59).

Please, consider reviewing the code and telling me what to fix before the code can be accepted. I realize that there is quite a lot of changes and it would be better if the changes were split into many smaller PRs. But I struggled a bit with Git. This PR should at least be without any conflict.


category_encoders/__init__.py
```diff
@@ -31,5 +32,6 @@
     'PolynomialEncoder',
     'BaseNEncoder',
     'LeaveOneOutEncoder',
-    'TargetEncoder'
+    'TargetEncoder',
+    'WoeEncoder'
```
wdm0006: Should we capitalize WOE since it's an acronym? I'm not sure there is a right or wrong, but whichever we go with here we should remember to keep consistent in future classes.

category_encoders/woe.py
```diff
@@ -0,0 +1,290 @@
...
+            Returns self.
+
+        """
+
+        # Unite parameters into pandas types
+        X = convert_input(X)
+        y = pd.Series(y, name='target')
+
+        # The lengths must be equal
+        assert X.shape[0] == y.shape[0]
```
wdm0006: So this is another case of pretty subjective note: since asserts can be turned off globally by users, I tend to assume that they will be. So if it's important to check something, and its not internally invariant (these are checking user input), we should check them without an assert, and raise a value error. Otherwise we can't assume that these checks ever run.

janmotl: Agreed. Changed all asserts into exceptions.

category_encoders/__init__.py
janmotl: PEP 8 says WOE -> changed.

wdm0006:
This looks good to merge, go ahead and merge it in if you're done with changes.

PR #113, opened 2018-08-28
Issue 109 smoothing min sample

JohnnyC08:
Here is the PR to address https://github.com/scikit-learn-contrib/categorical-encoding/issues/109

The fix was to pass `smoothing` and `min_samples_leaf` into the `target_encode` function for the fit function. 

I had some trouble finding test methods that worked for python3 and 2. 

I also found an issue where the calculation in python 2 was incorrect because the `-(n-k)/f` division was integer division leaving off the decimal. So I cast the smoothing to a float which fixed the python 2 issue.

Also, we were recalculating the means in the map when we already did so at the pandas level, so I removed that too as it was also causing python 2 problems.

Let me know what you think.

### DIFF
--- a/category_encoders/target_encoder.py
+++ b/category_encoders/target_encoder.py
@@ -119,7 +119,9 @@ class TargetEncoder(BaseEstimator, TransformerMixin):
             mapping=self.mapping,
             cols=self.cols,
             impute_missing=self.impute_missing,
-            handle_unknown=self.handle_unknown
+            handle_unknown=self.handle_unknown,
+            smoothing_in=self.smoothing,
+            min_samples_leaf=self.min_samples_leaf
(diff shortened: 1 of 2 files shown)


janmotl:
Good catches. I would just move the cast from the loop into the constructor.

JohnnyC08:
@janmotl Done

janmotl:
@JohnnyC08 Thank you.

JohnnyC08:
@janmotl No problem. Thanks for the opportunity to work on it your very nice library I like it. I have some more questions about this specific implementation versus what's in the initial target encoding paper. For now I'll close the related issue 109 and then later open up a new issue to ask my questions.

Thanks again!

PR #119, opened 2018-09-06
Issue 106 retrain leave one out

JohnnyC08:
Addresses https://github.com/scikit-learn-contrib/categorical-encoding/issues/106

To make the LeaveOneEncoder retrainable we follow @janmotl suggestion and split the `leave_one_out` method into methods for fit and transform respectively.

Even once that is done, the refit would fail because fit would set `self.cols` if `cols` is not passed into the constructor. To alleviate that problem, the `cols` from constructor get set to `self.static_cols` and in the fit method we set `self.cols` to `self.static_cols` if it is not none. If `self.static_cols` is None, then set `self.cols` to `get_obj_cols(X)`. So, we can retrain the encoder without specifying the columns in the constructor when each new DataFrame has different columns. That was my solution, please tell me what you think.

### DIFF
--- a/category_encoders/leave_one_out.py
+++ b/category_encoders/leave_one_out.py
@@ -80,7 +80,7 @@ class LeaveOneOutEncoder(BaseEstimator, TransformerMixin):
         self.drop_invariant = drop_invariant
         self.drop_cols = []
         self.verbose = verbose
-        self.static_cols = cols
+        self.use_default_cols = cols is None  # important when we call fit() repeatedly
         self.cols = cols
         self._dim = None
         self.mapping = None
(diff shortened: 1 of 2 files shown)


janmotl:
I even didn't realize that we will need a flag to see whether to use the implicit `cols` or the passed `cols`. Good work.

I would just consider changing the type of `static_cols` into a boolean and give it some comment:
```python
self.use_default_cols = cols is None # important when we call fit() repeatedly
```

Out of curiosity, I checked other encoders. And they all fail when the column names get changed.

JohnnyC08:
@janmotl Made the requested change.

Want to make a new issue to track what you found against the other encoders? I'm sure we can write an abstraction, maybe an abstract class, so that we will have to only write this logic in one place. We can even have the abstract class require separate methods for creating the mapping for fit and use it for transforming similar to what we did here.

janmotl:
I made a new issue, which contains a parameterized test.

TargetEncoder and WOEEncoder can be modified like LeaveOneOutEncoder. And that is where I would start. The remainder of encoders is structured differently and respective solutions may differ.

PR #128, opened 2018-09-29
Add a fit method for helmert encoding that builds a mapping and then …

JohnnyC08:
Addresses https://github.com/scikit-learn-contrib/categorical-encoding/issues/91.

To preserve dimensions we add fit functionality which stores dataframes indexed by the ordinal which represents the category. To handle the case where new values are passed we add a zero index to each dataframe that represents the unknown case. During the apply, we replicate the one hot encoder functionality and add the new columns and then reindex the data frame.

@janmotl Thank you for taking the time to explain helmert encoding to me. Once I started to get it I felt dumb for not seeing it in the first place. I used your tests and everything seems to be working.

 I marked as WIP so we can sharpen the the Helmert encoding case before tackling BackwardDifferenceEncoder, PolynomialEncoder and SumEncoder.


janmotl:
I am not sure that it is an issue or whether it is relevant, but following test does not pass:
```python
    def test_helmert_2cols(self):
        train = [['A', 'A'], ['B', 'B'], ['C', 'C']]

        encoder = encoders.HelmertEncoder()
        encoder.fit(train)
        obtained = encoder.transform(train)

        expected = [[1, -1, -1, -1, -1],
                    [1,  1, -1,  1, -1],
                    [1,  0,  2,  0,  2]]
        self.assertEqual(obtained.values.tolist(), expected)
```
Currently, HelmertEncoder returns an "intercept" column for each encoded feature. That is unnecessary. Sure enough, `drop_invariant` argument can take care of that. But it also removes the first "intercept" column that some people may expect/need.

For reference, `patsy` seems to generate only one intercept per dataset, as expected:
```python
import pandas as pd
from patsy import dmatrices

data = pd.DataFrame([['A', 'A', 3], ['B', 'B', 3], ['C', 'C', 3]], columns=['a', 'b', 'y'])
outcome, predictors = dmatrices("y ~ a + b", data)
print(predictors)
```

Btw., just for my curiosity, what is the reason that we first perform ordinal encoding? Missing value treatment?

JohnnyC08:
I went ahead and made it so only one intercept is returned if there are multiple categorical columns being processed. 

As for the use of the ordinal encoder, we don't necessarily need it for the initial encoding, but it makes the missing value case much simpler.

I understand what you meant about the multiple intercept columns. In your example we had one because the formula interrelated the two variables and in our case we have independent encodings, so does it make sense to have multiple intercepts? For now, I made it only one and we can go from there.

The next thing I want your opinion on is column ordering. So it seems right now the new columns get placed as the first columns in the transformed data frame. So, we can make it so the column to get transformed gets replaced by the new feature columns, but if we're transforming multiple categorical variables and we have one intercept column, then where should we put the intercept column.

janmotl:
When we have 2 independent columns, we still should have just one intercept:
```python
import pandas as pd
from patsy import dmatrices

data = pd.DataFrame([['A', 'X', 3], ['B', 'Y', 3], ['C', 'Z', 3]], columns=['a', 'b', 'y'])
outcome, predictors = dmatrices("y ~ a + b", data)
print(predictors)
```
If nothing else, regression models would dislike multiple intercepts because of their colinearity.

When it comes to order of columns, I propose to preserve the order of the columns in the input. And place the intercept as the first column in the output dataframe. At least this is the convention used by regression models as implemented in R and H2O.

Another edge scenario is what to do when the input does not contain anything to transform (e.g.: it contains only numerical columns or no columns at all). Should we add the intercept? I say yes, we should. The downstream model may expect the presence of the intercept. Plus, it makes the code tiny bit nicer.

JohnnyC08:
Ok, gonna tackle the `BackwardDifferenceEncoder` now. I'm going to make a new test class because test_encoders is getting pretty big

janmotl:
I agree with the split of `test_encoders`. Possibly all encoder specific tests could be moved into `test_backward_difference`, `test_basen`, ..., `test_target_encoder` based the name of the original file that they are testing.

JohnnyC08:
Ok I'm going through the other encoders right now. I'll split out the helmert one too and ensure the filenames match the pattern you described.

JohnnyC08:
@janmotl 
I've went ahead and have taken care of the BackwardDifferenceEncoder, PolynomialEncoder and SumEncoder as well. 

I have split tests for helmert, backwards difference, polynomial, and sum encoder into their own test class. I did not split out the target encoder tests in this PR to minimize the amount of files modified to make review easier. Also, it's unrelated to the task at hand, so a new PR is more appropriate. 

I did a search of the project for dmatrix and it seems we caught all the encoders.

Are there any other edge cases we should consider or are they any changes you'd like me to make?

janmotl:
@JohnnyC08 Good work. If you want to, go ahead with moving the encoder specific tests into separate files.

PR #135, opened 2018-10-16
Speed up TargetEncoder with vectorization.

jkleint:
Roughly 300X faster and uses less memory.
`fit()` computes category smoothed means with vector ops, stores mapping as (dict of) Series.
`transform()` then just calls Series.map(mapping) to do vectorized lookup with the mapping Series.

Also fixes IndexError when passing DataFrames with non-contiguous indexes.

Also fixes not being able to re-use TargetEncoder by always setting `mapping=None` in `fit()` (you don't want fit to remember anything from previous calls).

```
rows, cats = 1000000, 1000
X = pd.DataFrame({'x': np.random.randint(0, cats, rows).astype(str)})
y = pd.Series(np.random.rand(rows))

# old
%time TargetEncoder().fit_transform(X, y)
CPU times: user 1min 24s, sys: 71.2 ms, total: 1min 24s
Wall time: 1min 24s

# new
%time TargetEncoder().fit_transform(X, y)
CPU times: user 235 ms, sys: 43.6 ms, total: 279 ms
Wall time: 276 ms
```

### DIFF
--- a/category_encoders/target_encoder.py
+++ b/category_encoders/target_encoder.py
@@ -3,7 +3,6 @@ import numpy as np
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
 from category_encoders.utils import get_obj_cols, convert_input, get_generated_cols
-from sklearn.utils.random import check_random_state
 
 __author__ = 'chappers'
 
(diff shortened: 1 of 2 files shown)


janmotl:
Hi @jkleint, couldn't you look at LeaveOneOutEncoder and possibly vectorize it as well? I failed in the attempt.

jkleint:
Sure, I'm happy to take a look at it.

PR #139, opened 2018-10-18
Implemented get_feature_names() for OneHot and Hashing transformers.

datarian:
In reply to https://github.com/scikit-learn-contrib/categorical-encoding/issues/79.

I have implemented get_feature_names for OneHotEncoder and HashingEncoder as an example of how it could be done. Happy to get feedback before continuing on the remaining encoders.

I use both personally and find them to work well.

### DIFF
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -73,6 +73,7 @@ class HashingEncoder(BaseEstimator, TransformerMixin):
     Large Scale Multitask Learning. Proc. ICML.
 
     """
+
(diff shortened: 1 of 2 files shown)


datarian:
Too many issues. Will continue working on this.

janmotl:
Don't give up. There are just 2 issues that appear repeatedly:
1. When you train OneHotEncoder with `drop_invariant=False`, `generated_cols` is not defined.
2. `n_components` in HashingEncoder is not iterable.

datarian:
Yeah, I won't, don't worry. Just need a bit more time for a thorough revamp. Will PR again soon ;)

PR #140, opened 2018-10-18
Categorical handling

JohnnyC08:
Addresses https://github.com/scikit-learn-contrib/categorical-encoding/issues/86

@janmotl Here is my approach to setting up categorical datatype compatibility.

I've also started storing the datatype of the column so that way the inverse transform can set the types correctly. Luckily, there were not a lot of core things that had to change.

I was also able to ensure the ordinal encoder set it's values by the order of the categories in the data type.

Tell me what you think.


janmotl:
Looks good. Thank you.

PR #141, opened 2018-10-19
Get feature names

datarian:
Implemented get_feature_names for HashingEncoder, OneHotEncoder and OrdinalEncoder.

For my purposes, these work now. Not fully tested. It's more of a proposal for a concept. If liked, I will gladly implement for the rest of the encoders, incorporating any feedback.


janmotl:
Awesome! Before merging:
1. The PR must be without conflicts - the code in the master has, unfortunately, changed in the meantime.
2. If possible, the PR should not alter formatting (like line breaks and spaces) of the code - it makes review unnecessarily time consuming.
3. I think `# Indicate no transformation has been applied yet` in OHE is a residue from some previous code.
4. If we call `get_feature_names()` before `transform()`, the exception is not raised.
5. Typo in `...the list of of new...`

janmotl:
I have messed up the PR by deleting files (I just wanted to remove them from the PR). It's possibly the best to make a new PR, which is not going to reformat the code and contains only files that it should contain.

Nevertheless:
1. When HashEncoder encodes 2 or more columns, `get_feature_names` returns wrong results:
```python
    def test_get_feature_names(self):
        X = pd.DataFrame([['A','B','C']], columns=['col1', 'col2', 'col3'])
        enc = encoders.HashingEncoder(cols=['col2', 'col3'])
        result = enc.fit_transform(X)
        obtained = enc.get_feature_names()
        expected = set(result.columns) - set('col1')

        self.assertEqual(len(expected), len(obtained))
```
2. Beware that in Python 2 set does not preserve order of insertion.
3. Instead of using `is_fitted` you may test for `self.feature_names is None`. But that is a detail.

datarian:
> 
> 
> I have messed up the PR by deleting files (I just wanted to remove them from the PR). It's possibly the best to make a new PR, which is not going to reformat the code and contains only files that it should contain.
> 
> Nevertheless:
> 
>     1. When HashEncoder encodes 2 or more columns, `get_feature_names` returns wrong results:
> 
> 
> ```python
>     def test_get_feature_names(self):
>         X = pd.DataFrame([['A','B','C']], columns=['col1', 'col2', 'col3'])
>         enc = encoders.HashingEncoder(cols=['col2', 'col3'])
>         result = enc.fit_transform(X)
>         obtained = enc.get_feature_names()
>         expected = set(result.columns) - set('col1')
> 
>         self.assertEqual(len(expected), len(obtained))
> ```
> 
>     1. Beware that in Python 2 set does not preserve order of insertion.
> 
>     2. Instead of using `is_fitted` you may test for `self.feature_names is None`. But that is a detail.

Sorry for the delay, I am really busy and working on this when time permits. Maybe I got hash encoding wrong: So it appears that when you pass in >1 column, these columns all get mangled into the same 8 - or whatever the precision is - hash columns that get created in hash_fn? Is that right? Because I assumed that for each col, a set of 8 hashing cols gets created.

janmotl:
Yes, all the features get mangled into the same 8 - or whatever the precision is - hash columns, regardless of the count of the input features. You may look at it as a (lossy) dimensional reduction, which works on categorical features.

datarian:
I have taken up work on this again and have ironed out the previous problems, and written unit tests for everything.

There is one thing I am not sure about in the existing behavior of the OneHot encoder that I'd like to clarify. It has nothing to do with `get_feature_names()`, I'm just wondering:

If no column names to transform are passed, (and hence the transformer selects all object / category columns), the DF that's returned includes numerical columns found in the original df. Is this intended? From my understanding, transformers usually return only columns that were either transformed or newly created. 

I use the df X built with test_utils.py for my unit test. The 3 numerical columns are prepended as can be seen in the assertion error (which I got when using `utils.get_generated_cols()`:
`AssertionError: Lists differ: ['unique_str_1', 'unique_str_2', 'unique_st[1912 chars]_-1'] != ['float', 'float_edge', 'unique_int', 'uniq[1954 chars]_-1']`


For now I get the column names by assigning `X_trans = self.transform(X)` in `fit()`, and taking the column names from that, so the feature_names are correct for the DF that's returned.

janmotl:
Small things:
1. `self.assertEquals()` -> `self.assertEqual()`
2. `import category_encoders.utils as ut` in `test_one_hot` is unused

> I have thought about doing so but I have the intuition that it would be preferable to have the feature names available after a call to fit(), right?

If Scikit is happy with getting the feature names only after transform() is called, I would pick the way of the least resistance.

datarian:
Thanks for the feedback. Now there's just the one test that fails. I tried finding out what's going on with that but to no avail so far.

datarian:
> If Scikit is happy with getting the feature names only after transform() is called, I would pick the way of the least resistance.

Judging from here: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/preprocessing/_encoders.py#L694, they require the transformer to be just fitted, not transformed.

janmotl:
Ok. It looks like OHE does not pass:
```python
    def test_get_feature_names_after_transform(self):
                enc = getattr(encoders, encoder_name)
                enc.fit(X, y)
                out = enc.transform(X_t)
                self.assertEqual(set(enc.get_feature_names()), set(out.columns))
```

datarian:
> 
> 
> Ok. It looks like OHE does not pass:
> 
> ```python
>     def test_get_feature_names_after_transform(self):
>         for encoder_name in encoders.__all__:
>             with self.subTest(encoder_name=encoder_name):
>                 enc = getattr(encoders, encoder_name)()
>                 enc.fit(X, y)
>                 out = enc.transform(X_t)
>                 self.assertEqual(set(enc.get_feature_names()), set(out.columns))
> ```

This is fixed now.

datarian:
You're welcome @janmotl 
Will stick around and look for issues I could help with.

janmotl:
@datarian That would be greatly appreciated :)

Issue #132, opened 2018-10-13
OneHotEncoder incorrectly encodes -1 when use_cat_names is True


jvs:
When `use_cat_names` is True, OneHotEncoder incorrectly encodes values that equal -1.

Here's a minimal example:
```python
from category_encoders.one_hot import OneHotEncoder

encoder1 = OneHotEncoder(cols=[0], use_cat_names=False)
encoder2 = OneHotEncoder(cols=[0], use_cat_names=True)

a = encoder1.fit_transform([[-1]])
b = encoder2.fit_transform([[-1]])

print('a =', a.values)
print('b =', b.values)
```

This outputs:
```
a = [[1 0]]
b = [[0 0]]
```

I would expect `b` to have the same values as `a`.

janmotl:
Thank you. The treatment of missing values, which should be represented by -1, is truly broken.

janmotl:
There is a clash of generated names in `get_dummies()`. 

Specifically, when we have all the following:
1. `handle_unknown == 'impute'`
2. `use_cat_names == True` 
3. the input contains `-1`

Then:
1. one column name with `-1` suffix gets created because `handle_unknown == 'impute'` 
2. and another column name with `-1` suffix gets created because `use_cat_names == True` and the input contains `-1`

And DataFrame, quite understandably, does not support two columns with the same name.

Possible solutions:
1. find a sentinel value that can never appear in the input data and use it to mark imputed columns
2. write a function that generates unique column names (e.g.: We may keep adding '#' to the end of the name, until we generate a name that has not been observed before. This can be implemented with a reasonable overhead by using a set to keep track of the seen names.)

I am not particularly happy with either approach. Some opinions?

janmotl:
I was wrong. It is possible to create a DataFrame with duplicate column names:
```python
    def test_one_hot_cat_names(self):
        # currently both these combinations will result into columns named 'Apple_A_1'
        X = pd.DataFrame({'Apple_A': ['1'], 'Apple': ['A_1']})
        enc = encoders.OneHotEncoder(use_cat_names=True)
        result = enc.fit_transform(X)

        self.assertEqual(len(result.columns.unique()), len(result.columns), 'Columns must have unique names')
```
It's just inconvenient to use the result (e.g.: indexing by name does not work). 

So, I see two options:
1. return a DataFrame with duplicate column names
2. write a function, which generates unique names

I prefer the second approach. Go ahead.

janmotl:
[...]

PR #145, opened 2018-10-21
To solve the misisng column name uniqueness case we concatenate the c…

JohnnyC08:
Addresses https://github.com/scikit-learn-contrib/categorical-encoding/issues/132

We maintain a map of the number of times we see the raw column and append the `#` the number of times we have seen that column. By using a map and updating and using it one loop through everything, we hope we don't add to the overall processing time.

### DIFF
--- a/category_encoders/one_hot.py
+++ b/category_encoders/one_hot.py
@@ -274,8 +274,7 @@ class OneHotEncoder(BaseEstimator, TransformerMixin):
                 if self.use_cat_names:
                     n_col_name = str(col) + '_%s' % (cat_name, )
                     found_count = found_column_counts.get(n_col_name, 0)
-                    if found_count == 0:
-                        found_column_counts[n_col_name] = 1
+                    found_column_counts[n_col_name] = found_count + 1
                     n_col_name += '#' * found_count
                 else:
                     n_col_name = str(col) + '_%s' % (class_, )
--- a/category_encoders/one_hot.py
+++ b/category_encoders/one_hot.py
@@ -30,8 +30,8 @@ class OneHotEncoder(BaseEstimator, TransformerMixin):
         impute is used, an extra column will be added in if the transform matrix has unknown categories. This can cause
         unexpected changes in the dimension in some cases.
     use_cat_names: bool
-        if True, category values will be included in the encoded column names. Otherwise category
-        indices will be used.
+        if True, category values will be included in the encoded column names. Since this can result into duplicate column names, duplicates are suffixed with '#' symbol until a unique name is generated.
+        If False, category indices will be used instead of the category values.
 
     Example
     -------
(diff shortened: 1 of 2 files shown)


janmotl:
Beware that we can still get a clash of names:
```python
    def test_one_hot_duplicate_names(self):
        X = pd.DataFrame({'sea_rch': ['dummy', 'found'],
                          'sea': ['rch#-1', 'found']})
        encoder = encoders.OneHotEncoder(use_cat_names=True)

        result = encoder.fit_transform(X)
        columns = result.columns.tolist()
        
        self.assertEqual(len(set(columns)), len(columns))
```

A citation that came to my mind:
> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

What we are trying to deal with is an old problem. In Lisp they have gensym. Rust has raw strings with a variable count of `#`... 

I do not think we can find a sentinel value that can't appear in the input data but can be used in the column names simply because both, input data and the output can be strings. Hence, the domain of the symbols in the input can be the same as the domain of the symbols in the output -> no sentinel value may exist (assuming independence of the symbols).

Requirements for the naming function:
1. It must be reasonably fast.
2. The behavior of the function must be described in the documentation. E.g.:
```
    use_cat_names: bool
        if True, category values will be included in the encoded column names. Since this can result into duplicate column names, duplicates are suffixed with '#' symbol until a unique name is generated. If False, category indices will be used instead of the category values.
```

But I do not care about which symbol(s) we use and where do we place them.

Preservation of the attribute order should be another issue. #132 is already nontrivial.

JohnnyC08:
Dang it you're right. Ok, I'm going to focus on the suffix `#` approach

JohnnyC08:
@janmotl 
Went ahead and went with the suffix approach. Tell me what you think. Thanks buddy.

JohnnyC08:
Ok, I have fixed the inverse transforming by creating a mapping object that gets used in get_dummies and reverse_dummies. That way we don't generate the column names at transform time, we generate them at fit time which I think makes sense. 

I tried to find opportunities to use the vectorization transforms we've started using in target encoder and the woe encoder, but I had trouble seeing the opportunities. Can you spot where I can use them?

Tell me what you think.

janmotl:
I am not sure this is going to be faster, but for each nominal column:
1. Let's have a dictionary from category -> OHE column index. Pandas Series can be used for that as well.
2. Let's prealocate output matrix with zeros.
3. Then we can just loop over the nominal column once, read the category value, look up the index, write at the appropriate row and column index `1`. This step can be vectorized.
4. Append the generated matrix (which is just for this column) to the result.

Alternative: Instead of having the dictionary, we can have an encoding matrix. And instead of writing ones into the prealocated table, we could write whole rows from the encoding matrix.

Either way, the advantage is that we go thru the nominal column only once and not as many times as there is unique values in the column.

But before doing something, it could be meaningful to benchmark other implementations like `pandas.get_dummies` and see how they do it.

PR #148, opened 2018-10-27
Speed up ordinal Encoder

JohnnyC08:
To speed up Ordinal Encoder we create a dictionary of column to series and use the series map in the transform step.

```python
s = pd.Series(np.random.choice(['a','b','c'], 100000))
s = pd.DataFrame({'letter': s})
```

pandas factorize
```python
a = pd.factorize(s['letter'])
CPU times: user 6.47 ms, sys: 1.61 ms, total: 8.08 ms
Wall time: 6.58 ms
```

old code
```python
b = ce.OrdinalEncoder().fit_transform(s)
CPU times: user 50.9 ms, sys: 2.04 ms, total: 53 ms
Wall time: 51.4 ms
```

new code
```python
c = ce.OrdinalEncoder().fit_transform(s)
CPU times: user 15.4 ms, sys: 2.04 ms, total: 17.5 ms
Wall time: 15.6 ms
```

I also had to work on the inverse transforms with the new series approach. I did not benchmark that at this time.

### DIFF
--- a/category_encoders/tests/test_ordinal.py
+++ b/category_encoders/tests/test_ordinal.py
@@ -52,7 +52,7 @@ class TestOrdinalEncoder(TestCase):
         a = encoder.transform(data)
         self.assertEqual(a.values[0, 1], 0)
         self.assertEqual(a.values[1, 1], 1)
-        pd.factorize
+
         encoder = encoders.OrdinalEncoder(impute_missing=False)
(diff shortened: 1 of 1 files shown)


JohnnyC08:
Yup, the SumEncoder would not work with `get_values` when calling `fit_without_intercept`. It was the only one too.

PR #165, opened 2019-01-05
WIP: Fixes For 

JohnnyC08:
### Things Done
1. Fixed documentation for multi-column output encoders
2. Use reindex instead of loc in those encoders to remove pandas warnings
3. Add test for invariant for all encoders
4. Fix documentation indentation for OrdinalEncoder.

### To-Do
1. BaseN AUC difference. 
2. Solve get feature names increasing runtime for Hashing Encoder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,21 @@
+v2.0.0
+======
+
+ * Added James-Stein, CatBoost and m-estimate encoders
+ * Added get_feature_names method
+ * Refactored treatment of missing and new values 
+ * Speed up the encoders with vectorization
+ * Improved compatibility with Pandas Series and Numpy Arrays
+
+v1.3.0
+======
+
+ * Added Weight of Evidence encoder
+
 v1.2.8
 ======
 
- * critical bugfix in hashing encoder
+ * Critical bugfix in hashing encoder
 
 v1.2.7
 ======
--- a/README.md
+++ b/README.md
@@ -22,7 +22,9 @@ Encoding Methods
  * Binary [5]
  * Hashing [1]
  * Helmert Contrast [2][3]
+ * James-Stein Estimator [9]
  * LeaveOneOut [4]
+ * M-estimator [7]
  * Ordinal [2][3]
  * One-Hot [2][3]
  * Polynomial Contrast [2][3]
@@ -30,13 +32,12 @@ Encoding Methods
  * Target Encoding [7]
  * Weight of Evidence [8]
 
-Usage
+Installation
 -----
 
-The package by itself comes with a single module and an estimator. Before
-installing the module you will need `numpy`, `statsmodels`, and `scipy`.
+The package requires: `numpy`, `statsmodels`, and `scipy`.
 
-To install the module execute:
+To install the package, execute:
 
 ```shell
 $ python setup.py install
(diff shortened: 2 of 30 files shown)


janmotl:
> Fix documentation indentation for OrdinalEncoder. @janmotl When I looked at it I couldn't see the difference. Could you tell me what's wrong?

It turned out that a just single space is missing in front of `mapping: list of dict` :)

> Add test for invariant for Backward Difference Encoding and Helmert

I think that we should rather loop over all the encoders, not just these two. Or simply change:
```python
def create_dataset(n_rows=1000, extras=False, has_none=True):
    """
    Creates a dataset with some categorical variables
    """
    random.seed(2001)
    ds = [[
        random.random(),                                                                        # Floats
        random.choice([float('nan'), float('inf'), float('-inf'), -0, 0, 1, -1, math.pi]),      # Floats with edge scenarios
        row,                                                                                    # Unique integers
        str(row),                                                                               # Unique strings
        random.choice(['A', 'B']) if extras else 'A',                                           # Invariant in the training data
        random.choice(['A', 'B_b', 'C_c_c']),                                                   # Strings with underscores to test reverse_dummies()
        random.choice(['A', 'B', 'C', None]) if has_none else random.choice(['A', 'B', 'C']),   # None
        random.choice(['A', 'B', 'C', 'D']) if extras else random.choice(['A', 'B', 'C']),      # With a new string value
        random.choice([12, 43, -32]),                                                           # Number in the column name
        random.choice(['A', 'B', 'C']),                                                         # What is going to become the categorical column
    ] for row in range(n_rows)]

    df = pd.DataFrame(ds, columns=['float', 'float_edge', 'unique_int', 'unique_str', 'invariant', 'underscore', 'none', 'extra', 321, 'categorical'])
    df['categorical'] = pd.Categorical(df['categorical'], categories=['A', 'B', 'C'])
    return df
```
And it will get tested in `test_classification()`. I also added seed in order to make sure we always test on the same data -> the test results will be reproducible.

> BaseN AUC difference. @janmotl The results for binary are exactly the same even though we see a degradation in BaseN. The Binary Encoder is not just a wrapper around BaseN for base 2 which leads me to believe I should check the other bases. What base(s) did you use for the benchmark?

I see an improvement (the newer version has higher AUC). I measured it against:
https://github.com/scikit-learn-contrib/categorical-encoding/tree/5c1cce428592f25dc34ad6f5a2c49d2b14c53950 
There are two differences in the outputs:
1. Treatment of None values (it used to return a row of zeros).
2. The encoding of string columns is also different. I encoded the first testing fold of autos.arff with the default setting of the encoders (base=2) and they are different (it's not just that the columns are permutated): [old.csv.zip](https://github.com/scikit-learn-contrib/categorical-encoding/files/2729784/old.csv.zip), [new.csv.zip](https://github.com/scikit-learn-contrib/categorical-encoding/files/2729783/new.csv.zip). But this is not an easily reproducible detector of the differences.

The main issue is that I do not know with which commit the AUC improved. I will try to find it.

janmotl:
Ok, a bit simpler test, which reproduces the change:
```python
    def test_proof(self):
        X, y, _ = arff_loader.load('autos.arff')
        enc = encoders.BaseNEncoder()
        enc.fit(pd.DataFrame(X['make'][0:100]), y[0:100])
        enc.transform(pd.DataFrame(X['make'][101:200])).to_csv('proof_of_change.csv')
```
It's just necessary to import arff_loader. And edit it to load the data from internet, not from csv.

The change of encoding happened from November 20 to November 27.

JohnnyC08:
I'm looking at the `get_feature_names` for the hashing encoder and I noticed that the feature names get to all the columns in the output data frame even if some of those of the columns are numerical and would never be transformed by the encoder. Is this intentional? Or should `get_feature_names` just return the feature names the encoder would output?

janmotl:
Good point. We should possibly write a test which employs `sklearn.pipeline.FeatureUnion` and tests `get_feature_names()`, `fit()`, `fit_transform()` and `transform()`. 

But it can be a can of worms. For example `FeatureUnion` is not in the older versions of sklearn -> TravisCI may complaint. Also, `sklearn.pipeline` does not (or at least did not) allow `fit_transform(X, y)` but only allowed `fit_transform(X)` -> LeaveOneOut does not have to work correctly, as it encodes training data differently from the scoring data. The same applies to Gaussian noise in TargetEncoder.

PR #170, opened 2019-03-21
Refactoring of test_helper

datarian:
As discussed in [#169](https://github.com/scikit-learn-contrib/categorical-encoding/pull/169), renamed test_helper to helper and added unit tests for it.


janmotl:
We forgot to change `test_polynomial.py`.

PR #172, opened 2019-03-22
Fix column order in basen inverse_transform

datarian:
fixes column order for basen inverse transform


datarian:
BaseN is fixed. OneHot is not working yet.

Regarding a parameterized test:

There is already one for 'BaseNEncoder', 'BinaryEncoder', 'OneHotEncoder', 'OrdinalEncoder'. Are there other encoders that to inverse_transform? or what kind of additional test do you have in mind?

janmotl:
It looks like `test_inverse_transform` covers it.

PR #183, opened 2019-04-30
add python 3.7 with unpinned deps on travis

bollwyvl:
Congrats on 2.0! 

Over on https://github.com/conda-forge/category_encoders-feedstock/pull/12 I can get it to build, but had to skip a bunch of tests. Unfortunately, I'm just a hobbyist user of the library, and don't have a lot to contribute in terms of exploring possible solutions.

This PR adds an excursion for travis with py37 and all the deps unpinned to see if it uncovers anything that we can attribute to dependency drift.

Thanks again!


bollwyvl:
Those are pretty much the test fails I'm seeing over on the feedstock. I'm partially relieved, as it means it isn't a anaconda/conda-forge difference, but it can still be maddening to find real top pins for dependencies, absent any other insights. There might be some tooling that might help bisect the errors we're seeing...

janmotl:
Thank you for the report.

A workaround: It looks like upgrade of Pandas from 0.23.4 to 0.24.x is responsible for that...

@JohnnyC08, could you have a look at `test_encoders.test_handle_missing_return_nan_train`, which is failing on OrdinalEncoder?

janmotl:
In Pandas 0.24, they changed the evaluation of nans to None. It used to evaluate to True. Now it is False.

JohnnyC08:
@janmotl I have managed to reproduce and fix by adding
```
X[column] = X[column].fillna(np.nan)

```
Before the map call.

I have found more than the OrdinalEncoder is affected. I am going to think on a possible better solution because I am worried about this being a subpar implementation. I will also think about possibly adding `None` as something new in the mapping series. 

Is there anything in the pandas release notes about this change? I found it interesting that `np.nan` is not the same as `None` anymore.

I will work more on this tonight as I have a busy day at work. Expect more details soon.

janmotl:
@JohnnyC08 Thank you.

I only looked at: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html

Possibly relevant change is: 
> Optional Integer NA Support

Another issue is in CatBoost. Now `unseen_values` variable contains a different content than it used to. And it is because of the change in the equality comparison in Pandas.

JohnnyC08:
@janmotl Now that I think about the fix, it isn't so bad. When I make the changes for the new version of pandas where should I PR into? This PR? Master?

janmotl:
I propose to make a new PR.

bollwyvl:
Thanks for the fast response, all! I'm not married to this PR by any
means... But guess I, as a lazy repackager, would like to see something
equivalent landed at some point!

bollwyvl:
ooh, pretty green checkmarks! I also merged the conda-forge after setting a top `pandas` pin in and removing the test skips, so `2.0.0` is up over there.

With respect to edge cases, might there be a place for [hypothesis](https://hypothesis.readthedocs.io) in constructing classes of cases which can be tested ~~exhaustively~~ harder? This is great if there is an oracle, but harder if it is difficult to predict what the outcome should be.

janmotl:
I made a new release `2.1.0`, which takes into account metamorphic testing. Thank you for the suggestion.

PR #186, opened 2019-04-30
Made compatible with latest version of pandas by replacing different…

JohnnyC08:
Made compatible with latest version of pandas by replacing different null values with singular null definition of np.nan

@janmotl I didn't mess with the versions this time around because when I did I could tell I was gonna go down the rabbit hole so I reverted and just changed the code :)

So, I am definitely unsure about CatBoost. I made the tests pass with the latest pandas version, but I have a gut feeling there are more edge cases we're missing. So things to keep in mind.

Tell me what you think!

### DIFF
--- a/category_encoders/cat_boost.py
+++ b/category_encoders/cat_boost.py
@@ -272,7 +272,7 @@ class CatBoostEncoder(BaseEstimator, TransformerMixin):
             level_notunique = colmap['count'] > 1
 
             unique_train = colmap.index
-            unseen_values = pd.Series([x for x in X_in[col].unique() if x not in unique_train])
+            unseen_values = pd.Series([x for x in X[col].fillna(np.nan).unique() if x not in unique_train])
 
             is_nan = X_in[col].isnull()
             is_unknown_value = X_in[col].isin(unseen_values.dropna())
@@ -282,7 +282,7 @@ class CatBoostEncoder(BaseEstimator, TransformerMixin):
 
             if y is None:    # Replace level with its mean target; if level occurs only once, use global mean
                 level_means = ((colmap['sum'] + self._mean) / (colmap['count'] + 1)).where(level_notunique, self._mean)
-                X[col] = X[col].map(level_means)
+                X[col] = X[col].fillna(np.nan).map(level_means)
             else:
                 # Simulation of CatBoost implementation, which calculates leave-one-out on the fly.
                 # The nice thing about this is that it helps to prevent overfitting. The bad thing
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -287,6 +287,8 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
             mapping_out = mapping
             for switch in mapping:
                 column = switch.get('col')
+
+                X[column] = X[column].fillna(np.nan)
                 X[column] = X[column].map(switch['mapping'])
 
                 try:


janmotl:
If it is explained in the comment why we impute with `np.nan` (with a reference to Git discussion), I am OK with that.

Nevertheless, I have encountered an unexpected behaviour of HelmertEncoder. And it is possible that some other encoders do not behave correctly as well (but I can be wrong), as illustrated with:

```python
    def test_metamorphic(self):
        # When we only slightly alter the input data or an irrelevant argument, the output should remain unchanged.
        x1 = ['A', 'B', 'B']                                    # Baseline
        x2 = ['Apple', 'Banana', 'Banana']                      # Different strings, but with the same alphabetic ordering
        x3 = pd.DataFrame(data={'x': ['A', 'B', 'B']})          # DataFrame
        x4 = pd.Series(['A', 'B', 'B'], dtype='category')       # Series with category data type
        x5 = np.array(['A', 'B', 'B'])                          # Numpy
        x6 = [None, 'B', 'B']                                   # Missing value
        x7 = ['Z', 'Y', 'Y']                                    # Different strings, reversed alphabetic ordering (it works because we look at the order of appearance, not at alphabetic order)

        y = [1, 1, 0]
        
        for encoder_name in (set(encoders.__all__) - {'HashingEncoder'}):  # Hashing encoder is, by definition, not invariant to data changes 
                enc1 = getattr(encoders, encoder_name)()
                result1 = enc1.fit_transform(x1, y)

                enc2 = getattr(encoders, encoder_name)()
                result2 = enc2.fit_transform(x2, y)
                self.assertTrue(result1.equals(result2))

                enc3 = getattr(encoders, encoder_name)()
                result3 = enc3.fit_transform(x3, y)
                self.assertTrue((result1.values == result3.values).all())

                enc4 = getattr(encoders, encoder_name)()
                result4 = enc4.fit_transform(x4, y)
                self.assertTrue((result1.values == result4.values).all())

                enc5 = getattr(encoders, encoder_name)()
                result5 = enc5.fit_transform(x5, y)
                self.assertTrue((result1.values == result5.values).all())

                enc6 = getattr(encoders, encoder_name)()
                result6 = enc6.fit_transform(x6, y)
                self.assertTrue((result1.values == result6.values).all())  # Helmert fails here

                enc7 = getattr(encoders, encoder_name)()
                result7 = enc7.fit_transform(x7, y)
                self.assertTrue((result1.values == result7.values).all())


                # Arguments
                enc9 = getattr(encoders, encoder_name)(return_df=False)
                result9 = enc9.fit_transform(x1, y)
                self.assertTrue((result1.values == result9).all())

                enc10 = getattr(encoders, encoder_name)(verbose=True)
                result10 = enc10.fit_transform(x1, y)
                self.assertTrue((result1.values == result10.values).all())

                # Note: If the encoder does not support these arguments/argument values, it is OK/expected to fail.
                # Note: The indicator approach is not tested because it adds columns -> the encoders that support it are expected to fail.
                enc11 = getattr(encoders, encoder_name)(handle_unknown='return_nan', handle_missing='return_nan')  # Quite a few algorithms fail here because of handle_missing
                result11 = enc11.fit_transform(x1, y)
                self.assertTrue((result1.values == result11.values).all(), 'The data do not contain any missing or new value -> the result should be unchanged.')

                enc12 = getattr(encoders, encoder_name)(handle_unknown='value', handle_missing='value')  # Helmert fails here
                result12 = enc12.fit_transform(x1, y)
                self.assertTrue((result1.values == result12.values).all(), 'The data do not contain any missing or new value -> the result should be unchanged.')

                enc13 = getattr(encoders, encoder_name)(handle_unknown='error', handle_missing='error', cols=['x'])  # Quite a few algorithms fail here because of handle_missing
                result13 = enc13.fit_transform(x3, y)
                self.assertTrue((result1.values == result13.values).all(), 'The data do not contain any missing or new value -> the result should be unchanged.')
```

But it is possibly a new issue (it is related to nan handling but the issues are not Pandas version related).

JohnnyC08:
I will take a look. Thanks for pointing it out

janmotl:
After a bit of testing, it turned out that all encoders are affected. My resolution was to simply drop support for `None` and support only `NaN` onward.

PR #187, opened 2019-05-03
Count Encoder

JoshuaC3:
I have written a count encoder with various functionality. The idea here is so we can use a simple type of categorical encoding as a benchmark to compare the other types of encoding, from one library.

### DIFF
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -8,6 +8,7 @@
 
 from category_encoders.backward_difference import BackwardDifferenceEncoder
 from category_encoders.binary import BinaryEncoder
+from category_encoders.count import CountEncoder
 from category_encoders.hashing import HashingEncoder
 from category_encoders.helmert import HelmertEncoder
 from category_encoders.one_hot import OneHotEncoder
--- a/category_encoders/count.py
+++ b/category_encoders/count.py
@@ -1,0 +1,298 @@
+"""Count Encoder"""
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
+import category_encoders.utils as util
+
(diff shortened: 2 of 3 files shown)


janmotl:
Good work.

I think that:
`import category_encoders.tests.test_utils`
in the test file should be now:
`import category_encoders.tests.helpers`

And the helper methods should be modified to accept an optional argument `random_seed` with some default value.

What may turn out to be troublesome in the future is that if someone extends helper methods with new columns, it may affect the content of the current columns. And the tests will fail.

Travis CI should pass at least now.

---

Argument `impute_missing` in count encoder should be now `handle_missing` like in other encoders in categorical-encoding v2. 

---

Some typos in the doc:
"the minimun count" -> "the minimal count" 
"to long" -> "too long"
"In this case the" -> "In this case, the"
"delimeter" - > "delimiter"

---

A proposal:
```
        min_group_name: None, str
            Set the name of the combined minimum groups when the defaults become
            to long. Default None. In this case the category names will be joined
            with a `_` delimeter.
```
->
```
        min_group_name: None, str
            the name of the "leftover" group. If not set, the category names in the "leftover" 
            group get joined with a `_` delimiter and used as min_group_name. 
            Note: The default name can be long. And the name may keep changing, for example, 
            in cross-validation.
```

---

`fit_transform` is not necessary in unsupervised encoders - the definition of it in `TransformerMixin` is sufficient. I leave it up to you whether you like to keep it because of the customized docstring.

---

`count_encode` does not need `y`. I propose to remove it from the method signature.

JoshuaC3:
Hi, sorry, I realised I didn't do a pull before doing this so my repo was a good few months out of date. Silly mistake. I'll make another PR with the upto date test utils/helpers and copy across your comments and implement the changes.

PR #191, opened 2019-05-17
Count encoder with multiple options.

JoshuaC3:
1.  import category_encoders.tests.helpers addressed.

2. Helper methods should be modified to accept an optional argument random_seed with some default value.

3. Argument impute_missing is now handle_missing like in other encoders.

4. Some typos corrected in docs.

5. Note added to min_group_name as proposed
            Note: The default name can be long. And the name may keep changing, for example, 
            in cross-validation.

6. fit_transform removed for simplicity.

7. count_encode y removed

8. count_encode split into _count_encode_fit and count_encode_transform for better functionality and readability.

9. Improved test coverage of CountEncoder

10. Added check upon __init__

11. Added checks upon dict creation

12. Added dict functionality so individual column could be controlled

13. Added CountEncoder to readme with a reference.


janmotl:
It looks good. Thank you. Now we have to just appease TravisCI:
1. Can we replace `isna()` and `isnotna()` with `isnull()`, respectively with `~isnull()`? 
It looks to be [equivalent](https://datascience.stackexchange.com/questions/37878/difference-between-isna-and-isnull-in-pandas). And it would save us from the need to update the minimal requirements of Pandas.

2. Similarly, can we replace calls of this type:
```python
self.assertTrue(np.isin([28, 44], [44, 28, 423, 123]).all())
```
with something that does not require an update of Numpy:
```python
self.assertTrue(all([item in [44, 28, 423, 123] for item in [28, 44]]))
``` 
?

Reasoning: I like to update the minimal requirements only if the gain outweighs the disadvantages.

JoshuaC3:
1. and 2. addressed in #f3877f0

janmotl:
Good. Can we also make it to pass the tests with [Python 2.7](https://travis-ci.org/scikit-learn-contrib/categorical-encoding/jobs/534732647)?

JoshuaC3:
> Good. Can we also make it to pass the tests with [Python 2.7](https://travis-ci.org/scikit-learn-contrib/categorical-encoding/jobs/534732647)?

It looks like Python 2 uses a different random generator to Python 3, hence we get a different DataFrame with different counts etc. Working on this now.

janmotl:
Maybe it is better to test against a fixed testing dataset. Otherwise, we will be encountering this issue repeatedly. A side effect is that the dataset will be really small -> the tests will be fast and easy to reason about.

janmotl:
Congratulation for beating TravisCI!

JoshuaC3:
@janmotl great! Is there anything else I need to do, e.g. add to docs, etc

janmotl:
I don't think it is necessary to do anything more. Thank you for your contribution!

And feel free to make new submissions.

ekerazha:
@janmotl could the benchmarks (e.g. https://github.com/scikit-learn-contrib/categorical-encoding/tree/master/examples/benchmarking_large) be updated with the new models like this CountEncoder? Thanks

janmotl:
@ekerazha Done. You will find the result in `examples/benchmarking_large`.

Personally, I treat CountEncoder as a "feature extraction" tool rather than a standalone "encoder". I.e.: Whenever I apply CountEncoder, I also apply some other encoder and pass both results to the downstream model. Nevertheless, it looks like CountEncoder is good even alone.

PR #194, opened 2019-06-05
Add Multi-Process Supported HashingEncoder

SLLiu96:
Add an extended hashing encoder called NHashingEncoder. By using multi-process, it's several times faster than HashingEncoder. On i5-8259U, encoding 1,000,000 samples by HashingEncoder takes 720+ seconds while NHashingEncoder with parameter "max_process=4" only takes 230+ seconds. On 16x3.2GHz CPU + 64G Memory linux, encoding 20million samples by HashingEncoder takes over 4 hours while NHashingEncoder with parameter"max_process=8" only takes 20 minutes.


janmotl:
Nice.

1. The example in `NHashingEncoder` suggests that with the default `max_process=1` we get 3 fold acceleration. I guess that can be explained by missing `max_process=4` in the example.

2. Doesn't the backend has some function to autodetect the count of processors in the computer? If nothing else, numpy has it. We could set it as the default value.

3. What is the use case for `max_sample` argument? 

4. I would suggest to add parallelism into `HashingEncoder` in order to keep the class count low. Also add your name into __author__ field in `HashingEncoder`.

5. TravisCI must pass before I can merge the PR. I think that by merging `NHashingEncoder` with `HashingEncoder`, many errors will disappear.

SLLiu96:
@janmotl 
Appreciate it !!
Sorry I missed the max_process=4 when copying code from pycharm, will fix it next commit!

max_sample is for low-memory machine, it allows we custom the num of samples the each process get from dataset at each time. For example, as default, 4 Cores CPUs with 100000 samples will work as 25000 samples for each core, but we can give every core 10000 samples at each time after it finished pre-10000 samples by setting max_sample=10000.

Autodetect the count of processors will be contained in next commit! Thanks for suggestion!

I do prefer to add this inside HashingEncoder for clearer class struct, it separate because I'm not sure if this idea good or bad~

The CI error, well...working on it!

Once again, thanks a lot! I'd like to comment here and at you once I fix it if doesn't bother you!

SLLiu96:
@janmotl 
Don't know what's going wrong, I just add some codes into hashing.py
Could you please review this? Thanks a lot! I'm new to PR, sorry for bothering u!

janmotl:
I thought that we are just going to do something like:
1. rename the current methods like `transform()` in `HashingEncoder` into `_transform()`
2. copy paste method `transform()` from `NHashingEncoder` into `HashingEncoder` 
3. tell `transform()` to call `_transform()`

In the end, we would have just a single class `class HashingEncoder(BaseEstimator, TransformerMixin):` in `hashing.py`.

SLLiu96:
@janmotl 
Thank you so much!

But I'm a little concerned about the adaptation. I've done some test on my machine, the results of single-process hashing and multiple-process hashing are different, because the assemble of samples is different, we can consider it as an addition of samples. In my test, multi-process usually brings a lower logloss or higher scores than single-process, but I can't prove that it's certainly better at all time. 

By experience, we don't worry about new samples as long as the data after hashing has the same length. Sometimes I suspect that it perfomance better just because multi-process makes the data more random just lisk the random-forest usually better than decision tree!

So, I suggest put it as a function into HashingEncoder & name it as "n_transform()" and keep the current one for a while to make sure that it's right.

Hoping for your suggestion!

janmotl:
HashingEncoder should be deterministic, not stochastic. Hence, the results from the parallelized version should be identical to the serial version.

Check the ordering of the rows in the returned table - the rows could be permutated.

SLLiu96:
@janmotl 
Thank you!

Didn't find that I make it reset_id in the end, but still can't explain why multi-process get lower logloss, maybe my dataset's a special case!

Anyway! I'll folow your suggestions and add an auto-sort in the end!

Thanks again!

janmotl:
If 
1. the dataset from the serial encoder is equal the dataset from the parallel encoder
2. and both classifiers use the same seed, 

you should be getting the same logloss.

> is keeping sort important? Or maybe different sort is acceptable

With `return_df=False`, a different ordering of the samples could lead into nasty surprises, because `numpy.ndarray` doesn't preserve index -> the training data would not be aligned with the labels anymore.

SLLiu96:
@janmotl 
Thanks a lot!

Will check my code later about diff logloss.

Keep-sort and autodetect process are contained in, and autodetect process now gives max_process a default value.

All function inside HashingEncoder now, could you please review it again and tell me if anything's wrong (I strongly suspect that the "CI Check" will give me another error Org).

Appreciat it!

SLLiu96:
Seems I got another CI error now.

Don't know how to do, please help~~~~

janmotl:
I have executed all the tests locally. And HashingEncoder now fails in `TestEncoders.test_classification()`. Some other tests didn't finish:
```
ResourceWarning: subprocess 13075 is still running ResourceWarning, source=self)
```

My recommendation: get the class into the state that it passes all the tests locally. And only then push it to Git in order to check that the tests will pass with different library versions.

SLLiu96:
@janmotl 
Hi! Thanks for the suggestions!

Probably I made some mistakes in code, I've commited several times & got different error feedback, mostly caused by code. After fixing all of them, now I get this error:
```
Traceback (most recent call last):
  File "/home/travis/build/scikit-learn-contrib/categorical-encoding/category_encoders/tests/test_encoders.py", line 68, in test_classification
    self.assertTrue(isinstance(enc.transform(X_t), np.ndarray))
AssertionError: False is not true
```

In test_encoders.py, the code is:
```python
enc = getattr(encoders, encoder_name)(return_df=False)
enc.fit(X, np_y)
self.assertTrue(isinstance(enc.transform(X_t), np.ndarray))  # Error Here
```

It seems that, with parameter "return_df=False", the return value should be a np.ndarray but get another one.
But the code of controlling the type which to return seems OK
```python
if self.return_df or override_return_df:
    return data
else:
    return data.values
```

And when I test it myself like below:
```python
nhe = HashingEncoder(return_df=True)
nhe = nhe.fit(X, np_y)
return_data = nhe.transform(X_t)
print(type(return_data))

nhe = HashingEncoder(return_df=False)
nhe = nhe.fit(X, np_y)
return_data = nhe.transform(X_t)
print(type(return_data))
```

I got this log:
```
<class 'pandas.core.frame.DataFrame'>
<class 'numpy.ndarray'>
```

Same test code, same input, same hashing.py, but success in my test demo and failed in test_encoders.py, that's odd!

SLLiu96:
@janmotl 
Hello! 

Finally, all checks passed, also passed in local test!

I merged the new "transform()" & rename the old one to "_transform()" like you told me, and added some doc for detail. Maybe it's ready for merging, waiting for your review~

Thank you for helping me so much during the whole time!!!!

janmotl:
I will look at it on Wednesday.

Majority of the encoders should utilize inherent parallelization of numpy (which is used by pandas as the backend). You may check it by looking on core utilization on large datasets. A simple script that generates a dataframe with: columns=[encoder, average_core_utilization] could then placed into examples directory (e.g.: benchmarking_cpu). Or you can add the functionality into one of the benchmarks already present.

SLLiu96:
Thanks for suggestions!

I'll have a try. (Before that I may need to read the exist eamples)

SLLiu96:
Hi!

I added benchmarking of cpu utilization of encoders and placed them into examples folder, but not sure is that right.

Thanks for your review.

janmotl:
The benchmarking looks good. But it needs some text to explain what's going on. For example, what 100% utilization means (that all virtual cores are maxed?). It is also pity that we do not differentiate between training and scoring. And between average and peak utilization. 

I proposed some changes to the documentation in hashing.py. Change it further as you desire.

I am not sure it is wise to leave a commented code hanging there:
```python
        # if X is None:
        #     raise AttributeError("None data input")
```

Otherwise it looks good.

SLLiu96:
@janmotl 
That's a good suggestion, thanks!

Will update the benchmarking later! I may need to do some tests to make sure it's right before update, that may take a few days, but I'll try to add more.

Thanks for all your help!! I'll continue fixing bugs and updating.

SLLiu96:
@janmotl 
Hi! An update of benchmarking has been commited!

This time I Differentiate the peak and average of time(min time as peak) and cpu_utilization(max percentage as peak), also, every value of max_process are differentiated now, if other encoders supported by multi-process in future, maybe we can add them into it.

janmotl:
Good. I propose to restyle the text in the benchmark to:
```
Record the average and peak system-wide CPU utilization during encoder training and scoring.
The utilization is reported as a percentage and as such should always be in the range 0..100%.
E.g.: 50% means half of the logical CPUs are completely utilized in a multi-core device,
or half of the CPU is utilized in a single-core device.
```

> I tried to set the sample num larger like 100,000 but my machine made a lot of noise and heated soon, so I had to terminate it and set it smaller to 10,000.

I was thinking about that as well. A reasonable solution would be to create a large dataset and keep running each encoder on it for up to ~5 seconds. If it takes longer, just kill the encoder and move on. But of course, we can just explain in the documentation that if an encoder finishes in less than `cpu_sampling_rate`, its utilization is recorded as 0.

janmotl:
I tried to run `benchmarking_large` with the new HashingEncoder and it failed on deepcopy:
```
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 169, in deepcopy
    rv = reductor(4)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/synchronize.py", line 102, in __getstate__
    context.assert_spawning(self)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/context.py", line 356, in assert_spawning
    ' through inheritance' % type(obj).__name__
RuntimeError: Lock objects should only be shared between processes through inheritance
```
Is there some way how to get around that? I ask because deepcopy is also used in some parts of scikit (I am a bit surprised that we didn't encounter that already in scikit pipelines).

SLLiu96:
@janmotl 
That's a nice way, I will restyle the document and try to set a time counter, thanks!

I will check the error in benchmarking_large, and reply you later! I've seen this error several times, might know how to fix it. Usually it's my bad in code, may I ask that is the error happened in both python2 and python3 ?

SLLiu96:
@janmotl 
Sorry for that error!

I've update the HashingEncoder, the error happened mostly because of missing ''Manager()'' , I'm sorry I left it in my test code but didn't copy to the PR.

I successfully ran the benchmarking_large in my local test, would you please rerun that again and check if the error disappear now? Thanks for your review!

janmotl:
I am still getting errors:
```
Traceback (most recent call last):
  File "/Users/jan/Documents/Git/categorical-encoding/examples/benchmarking_large/benchmarking_large.py", line 114, in <module>
    folds, fit_encoder_time, score_encoder_time = train_encoder(X, y, fold_count, encoder)
  File "/Users/jan/Documents/Git/categorical-encoding/examples/benchmarking_large/util.py", line 28, in train_encoder
    encoder = deepcopy(encoder)  # Because of https://github.com/scikit-learn-contrib/categorical-encoding/issues/106
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 280, in _reconstruct
    state = deepcopy(state, memo)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 240, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/copy.py", line 161, in deepcopy
    y = copier(memo)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/managers.py", line 886, in __deepcopy__
    return self._getvalue()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/managers.py", line 817, in _getvalue
    return self._callmethod('#GETVALUE')
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/managers.py", line 811, in _callmethod
    raise convert_to_error(kind, result)
multiprocessing.managers.RemoteError: 
---------------------------------------------------------------------------
Unserializable message: Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/managers.py", line 289, in serve_client
    send(msg)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: can't pickle _thread.lock objects

---------------------------------------------------------------------------
```

My proposal: write a parameterized test in `test_encoders.py`, which executes deepcopy on each encoder.

SLLiu96:
@janmotl 
I'm checking this error, but actually, I wonder where the deepcopy executes.

Seems this error happened at line 114 in ``benchmarking_large.py`` , but the code is:

```python
print('Evaluating:', dataset_name, encoder.__class__.__name__, model.__class__.__name__)
```

However the traceback was telling us that the error happened at:

```python
scores, fit_model_time, score_model_time = train_model(folds, model)
```

Is the deepcopy executes at ``train_model()`` ? I checked online and found that is mostly because an object of custom class passed as a parameter when create a new process, however, there's no object of custom class:

```python
process = multiprocessing.Process(target=self.__require_data, args=(self.cols, thread_index + 1))
```

Sorry I don't know how to test the deepcopy, would you please give me an example of it?

janmotl:
I updated the benchmark today. Hence the row numbers are off.

>  would you please give me an example of it?
```python
from copy import deepcopy

    def test_deepcopy(self):
        # See: https://github.com/scikit-learn-contrib/categorical-encoding/pull/194
                enc = getattr(encoders, encoder_name)()
                enc2 = deepcopy(enc)
```

SLLiu96:
@janmotl 
Thanks so much for your example!

Sorry I'm quite new to unittest, didn't know much about how to create the test yet, but working on it.
Thanks for you example the error is located, I'm fixing it and will commit it as soon as I can!

I guess this's a tough one, probably need to modify some code in existing tests if I may, but I will let you know first.

SLLiu96:
@janmotl 
Hi! This commit contains deepcopy test now, and the error of deepcopy seems be fixed!
Thanks for your one more review!

I noticed that I commited so many times, is it necessary to close this PR and recreate one?

janmotl:
> deepcopy seems be fixed!

Yes. Large benchmark now works.

> I noticed that I commited so many times, is it necessary to close this PR and recreate one?

No. 

Do you want to do some more work on the benchmark/encoder? Or should I merge it now?

SLLiu96:
Thanks for review! I prefer to merge it if no other bugs~

I did consider to add the time counter in benchmarking_cpu, but after some tries I found that terminate after timeout dosen't work well in some encoders.

e.g. Some encoders' encoding time increases exponentially with the number of samples, or beacuse of the terminator needs to run in a sub-process, HashingEncoder with all-cores encoding, which is important in benchmarking, will be slowed down. Otherwise, some of the encoders are so much faster than some others, that makes the shape of initial dataset hard to define.

Anyway, I'm considering about another way to balance the encoding time, like auto-define the shape for each encoders before benchmarking, it's a feasible but not a good way, still finding a better one.

Thanks a loooot for helping me soooo much !

janmotl:
Ok. Thank you for the pull request.

SLLiu96:
Thank you as well !

PR #204, opened 2019-08-09
completed target encoding with smoothing

Ankur-singh:
Added a new categorical encoding technique, **target encoding with smoothing.** 

- [X] tested with numpy array & pandas dataframe as input

### DIFF
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -22,6 +22,7 @@ from category_encoders.woe import WOEEncoder
 from category_encoders.m_estimate import MEstimateEncoder
 from category_encoders.james_stein import JamesSteinEncoder
 from category_encoders.cat_boost import CatBoostEncoder
+from category_encoders.target_encoder_with_smoothing import TargetEncodingSmoothing
 
 __version__ = '2.0.0'
 
@@ -42,5 +43,6 @@ __all__ = [
     'WOEEncoder',
     'MEstimateEncoder',
     'JamesSteinEncoder',
-    'CatBoostEncoder'
+    'CatBoostEncoder',
+	'TargetEncodingSmoothing'
 ]
--- a/category_encoders/target_encoder_with_smoothing.py
+++ b/category_encoders/target_encoder_with_smoothing.py
@@ -1,0 +1,195 @@
+"""Target Encoder with smoothing"""
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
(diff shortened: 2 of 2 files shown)


janmotl:
Thank you. 

It looks like TargetEncoder also uses a sigmoid for smoothing (just the parameters are named differently). Can you compare TargetEncodingSmoothing with TargetEncoder?

Just by glance, we may handle differently: 

1. Unique values in the training data (we set them to be exactly equal to the global mean in order to decrease the overfitting)
2. The default setting of the sigmoid is different (I do not know which is generally better)
3. We unnecessarily check `y` in `transform()` (but it is quite important in LeaveOneOutEncoder)
4. You may handle string `y`.

I think we have a great opportunity to compare the implementations and fix deficiencies in each implementation. And eventually end up with a single implementation.

Quick notes:

1. It can be useful to run `test_encoders` to catch some evident differences
2. `y_` is not always defined
3. You may have to add `**params` as an argument to `__init__()` in order to silently ignore arguments like `verbose=True`,...

Ankur-singh:
@janmotl Thanks for the review. I am looking into it. Having a single implementation that would also handle smoothing will actually make more sense. ON IT!

Ankur-singh:
I came across [this ](https://maxhalford.github.io/blog/target-encoding-done-the-right-way/) blog on the internet. It says the following:
> There are other ways to do target encoding, such as this one which is rather popular on Kaggle. However it produces encoded variables which are very correlated with the output of additive smoothing, at the cost of requiring two parameters.

I guess, this pretty much solves the confusion.

janmotl:
To summarize it, there are at least 3 ways how to perform target encoding:
1. Take average
2. Use additive smoothing, which uses 1 parameter. This is implemented in MEstimateEncoder.
3. Use a sigmoid, which uses 2 parameters. This is implemented in TargetEncoder.

Overall, TargetEncoder tends to deliver slightly better results than MEstimateEncoder. See: https://github.com/scikit-learn-contrib/categorical-encoding/tree/master/examples/benchmarking_large/output. Of course, the results depend on the parametrization of the encoders. But these results are without tuning. And TargetEncoder still beats MEstimateEncoder even though it uses more parameters. 

If you are interested to contribute, it would be nice to add a support for poly-class targets in TargetEncoder - we currently support only binary targets. Reference for implementation: A preprocessing scheme for high-cardinality categorical attributes in classification and prediction problems by Daniele Micci-Barreca.

Ankur-singh:
I would for sure love to contribute. Thanks for the encouragement. I will go through the paper.

PR #220, opened 2019-10-30
Added support for 'y' with string data.

xKHUNx:
Previous version of CatBoostEncoder only works with numerical 'y', this update added support for string label for 'y'.

### DIFF
--- a/category_encoders/cat_boost.py
+++ b/category_encoders/cat_boost.py
@@ -5,6 +5,7 @@ import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
 import category_encoders.utils as util
 from sklearn.utils.random import check_random_state
+from sklearn.preprocessing import LabelEncoder
 
 __author__ = 'Jan Motl'
 
@@ -106,6 +107,7 @@ class CatBoostEncoder(BaseEstimator, TransformerMixin):
         self.sigma = sigma
         self.feature_names = None
         self.a = a
+        self.le_ = None
 
     def fit(self, X, y, **kwargs):
         """Fit encoder according to X and y.
@@ -129,6 +131,10 @@ class CatBoostEncoder(BaseEstimator, TransformerMixin):
 
         # unite the input into pandas types
         X = util.convert_input(X)
+        # encode the label into integers
+        self.le_ = LabelEncoder().fit(y)
+        y = self.le_.transform(y)
+        y = y.astype(int)
         y = util.convert_input_vector(y, X.index).astype(float)
 
         if X.shape[0] != y.shape[0]:
@@ -204,6 +210,10 @@ class CatBoostEncoder(BaseEstimator, TransformerMixin):
 
         # if we are encoding the training data, we have to check the target
         if y is not None:
+            # Encode labels from string to float (if not already float)
+            if y.dtype != np.float64:
+                y = self.le_.transform(y)
+
             y = util.convert_input_vector(y, X.index).astype(float)
             if X.shape[0] != y.shape[0]:
                 raise ValueError("The length of X is " + str(X.shape[0]) + " but length of y is " + str(y.shape[0]) + ".")


janmotl:
Awesome! A few notes:

1. It makes me nervous that some of the tests are failing (see the details from Travis). All unit tests in `category_encoders\tests` should pass on your computer.

2. When we are converting strings to numbers, we should possibly check whether there are exactly 2 unique values in `y` (and otherwise raise an error), because people may attempt to use this encoder on polynomial `y`, not realizing that it does not actually work properly.

3. We should add a new test into `test_cat_boost.py`, which tests the new support for the string `y`.

4. For ordinal encoding, we use the build in `category_encoders.OrdinalEncoder` instead of `sklearn.preprocessing.LabelEncoder`. After the switch, some of the errors will disappear (LabelEncoder is pickier about the format of the input).

janmotl:
I think that neither of the supervised encoders support multiclass labels because all supervised encoders always encode a single nominal feature into a single numerical feature. 

To support multiclass labels, you have to first OneHotEncode the label to generate binary labels. And then apply supervised encoders on the nominal features one time for each binary class. I.e.: When you have 3 nominal features and 5 classes, you end up with 3*5=15 numerical features.

Note: I assume that we are encoding _c_ classes into _c_ columns. But it is actually ok to encode the label just into _c-1_ columns.

janmotl:
I have implemented a simple wrapper for supervised encoders, which adds support for multi-class problems.

Test it and let me know if it worked for you.

```python
import copy
from category_encoders import utils
from sklearn.base import BaseEstimator, TransformerMixin
import category_encoders as encoders
import pandas as pd


# Multi-class wrapper: Extends supervised encoders to n-class labels, where n >= 2.
# The label can be numerical (e.g.: 0, 1, 2, 3,...,n), string or categorical (pandas.Categorical).
# Feature names are suffixed with underscore and the corresponding class name (edge scenarios like 'dog'+'cat_frog' vs. 'dog_cat'+'frog' are not handled).
class MultiClassWrapper(BaseEstimator, TransformerMixin):
    def __init__(self, feature_encoder):
        self.feature_encoder = feature_encoder
        self.feature_encoders = {}
        self.label_encoder = None

    def fit(self, X, y, **kwargs):
        # unite the input into pandas types
        X = utils.convert_input(X)
        y = utils.convert_input(y)
        y.columns = ['target']

        # apply one-hot-encoder on the label
        self.label_encoder = encoders.OneHotEncoder(handle_missing='error', handle_unknown='error', cols=['target'], drop_invariant=True)
        labels = self.label_encoder.fit_transform(y)
        labels.columns = [column[7:] for column in labels.columns]

        # train the feature encoders
        for class_name, label in labels.iteritems():
            self.feature_encoders[class_name] = copy.deepcopy(self.feature_encoder).fit(X, label)

    def transform(self, X):
        # unite the input into pandas types
        X = utils.convert_input(X)

        # transform the features
        features = pd.DataFrame()
        for class_name, feature_encoder in self.feature_encoders.items():
            df = feature_encoder.transform(X)
            df.columns = [str(column) + '_' + class_name for column in df.columns]
            features = pd.concat((features, df), axis=1)

        return features

    def fit_transform(self, X, y=None, **fit_params):
        # When we are training the feature encoders, we have to use fit_transform() method on the features.

        # unite the input into pandas types
        X = utils.convert_input(X)
        y = utils.convert_input(y)
        y.columns = ['target']

        # apply one-hot-encoder on the label
        self.label_encoder = encoders.OneHotEncoder(handle_missing='error', handle_unknown='error', cols=['target'], drop_invariant=True, use_cat_names=True)
        labels = self.label_encoder.fit_transform(y)
        labels.columns = [column[7:] for column in labels.columns]

        # fit_transform the feature encoders
        features = pd.DataFrame()
        for class_name, label in labels.iteritems():
            encoder = copy.deepcopy(self.feature_encoder)
            df = encoder.fit_transform(X, label)
            df.columns = [str(column) + '_' + class_name for column in df.columns]
            features = pd.concat((features, df), axis=1)
            self.feature_encoders[class_name] = encoder

        return features
```

Example usage:
```python
    def test_fit(self):
        x = np.array([
            ['a', 'b', 'c'],
            ['a', 'b', 'c'],
            ['b', 'b', 'c'],
            ['b', 'b', 'b'],
            ['b', 'b', 'b'],
            ['a', 'b', 'a'],
        ])
        y = [1, 2, 3, 1, 2, 2]
        encoder = encoders.TargetEncoder()
        wrapper = MultiClassWrapper(encoder)

        wrapper.fit(x, y)
        result = wrapper.transform(x)
        print(result)

    def test_fit_transform(self):
        x = pd.DataFrame([
            ['a', 'b', 'c'],
            ['a', 'b', 'c'],
            ['b', 'b', 'c'],
            ['b', 'b', 'b'],
            ['b', 'b', 'b'],
            ['a', 'b', 'a'],
        ], columns=['f1', 'f2', 'f3'])
        y = ['bee', 'cat', 'dog', 'dog', 'dog', 'dog']
        encoder = encoders.TargetEncoder()
        wrapper = MultiClassWrapper(encoder)

        result = wrapper.fit_transform(x, y)
        print(result)
```

xKHUNx:
Thanks for writing the wrapper! I have tested it and found some bugs:

1.  Inconsistencies with `self.label_encoder`,  where in `fit_transform` method, it has `use_cat_names=True` but not in `transform` method.
2. The `fit_transform` and `transform` method will return `n` copies of all features, for `n` is number of class labels, this includes non-categorical features, which is redundant.
3. Not really a bug, but the wrapper doesn't work optimally in binary class case (number of classes = 2). It will create 2 feature instead of just 1 feature for each categorical features, which is redundant.

I've fixed the issues listed above with the code provided below, here's the detail:

1.  Both methods now uses `use_cat_names=True`. Fixes issue (1).
2. The `fit_transform` and `transform` method will only return `n-1` copies of categorical features, for `n` is number of class labels. Numerical features is untouched. Fixes issue (2) and issue (3).

```
import copy
from category_encoders import utils
from sklearn.base import BaseEstimator, TransformerMixin
import category_encoders as encoders
import pandas as pd


# Multi-class wrapper: Extends supervised encoders to n-class labels, where n >= 2.
# The label can be numerical (e.g.: 0, 1, 2, 3,...,n), string or categorical (pandas.Categorical).
# Feature names are suffixed with underscore and the corresponding class name (edge scenarios like 'dog'+'cat_frog' vs. 'dog_cat'+'frog' are not handled).
class MultiClassWrapper(BaseEstimator, TransformerMixin):
    def __init__(self, feature_encoder):
        self.feature_encoder = feature_encoder
        self.feature_encoders = {}
        self.label_encoder = None

    def fit(self, X, y, **kwargs):
        # unite the input into pandas types
        X = utils.convert_input(X)
        y = utils.convert_input(y)
        y.columns = ['target']

        # apply one-hot-encoder on the label
        self.label_encoder = encoders.OneHotEncoder(handle_missing='error', handle_unknown='error', cols=['target'], drop_invariant=True, use_cat_names=True)
        labels = self.label_encoder.fit_transform(y)
        labels.columns = [column[7:] for column in labels.columns]
        
        self.n_classes = len(labels.columns)

        if self.n_classes > 2:
            # train the feature encoders
            for class_name, label in labels.iteritems():
                self.feature_encoders[class_name] = copy.deepcopy(self.feature_encoder).fit(X, label)
            # Remove last class to have a more compact encoding
            self.feature_encoders.pop(class_name)
        else:
            # binary class case
            self.feature_encoders = copy.deepcopy(self.feature_encoder).fit(X, y)

        return self

    def transform(self, X):
        # unite the input into pandas types
        X = utils.convert_input(X)
    
        if self.n_classes > 2:
            # transform the features
            features = pd.DataFrame()
            for class_name, feature_encoder in self.feature_encoders.items():
                df = feature_encoder.transform(X)
                # Select categorical features that are specified by encoder only
                new_feats = df[feature_encoder.cols]
                feat_names = [str(column) + '_' + class_name for column in df.columns if column in feature_encoder.cols]
                new_feats.columns = feat_names
                features = pd.concat((features, new_feats), axis=1)
            # Combine categorical and none categorical features as output
            features = pd.concat((df[df.columns[~df.columns.isin(feature_encoder.cols)]], features), axis=1)
        # binary class case    
        else:
            features = self.feature_encoders.transform(X)

        return features

    def fit_transform(self, X, y=None, **fit_params):
        # When we are training the feature encoders, we have to use fit_transform() method on the features.

        # unite the input into pandas types
        X = utils.convert_input(X)
        y = utils.convert_input(y)
        y.columns = ['target']

        features = self.fit(X, y).transform(X)

        return features
    
    def test_fit(self):
        x = np.array([
            ['a', 'b', 'c'],
            ['a', 'b', 'c'],
            ['b', 'b', 'c'],
            ['b', 'b', 'b'],
            ['b', 'b', 'b'],
            ['a', 'b', 'a'],
        ])
        y = [1, 2, 3, 1, 2, 2]
        encoder = encoders.TargetEncoder()
        wrapper = MultiClassWrapper(encoder)

        wrapper.fit(x, y)
        result = wrapper.transform(x)
        print(result)

    def test_fit_transform(self):
        x = pd.DataFrame([
            ['a', 'b', 'c'],
            ['a', 'b', 'c'],
            ['b', 'b', 'c'],
            ['b', 'b', 'b'],
            ['b', 'b', 'b'],
            ['a', 'b', 'a'],
        ], columns=['f1', 'f2', 'f3'])
        y = ['bee', 'cat', 'dog', 'dog', 'dog', 'dog']
        encoder = encoders.TargetEncoder()
        wrapper = MultiClassWrapper(encoder)

        result = wrapper.fit_transform(x, y)
        print(result)
```

janmotl:
Good points. I have uploaded the code into the master branch.

Notes: 
1. I am removing the single class before fitting the encoders (i.e. as soon as possible).
2. It is necessary to respect that `fit_transform(x, y)` behaves differently than `fit(x, y).transform(x)` in encoders like LeaveOneOut or CatBoostEncoder.
3. I did not incorporate a special path for binary case because I am leaving it up to the wrapper to deal with string labels.

xKHUNx:
Few questions:

> 1. I am removing the single class before fitting the encoders (i.e. as soon as possible).

What do you mean?

> 2. It is necessary to respect that `fit_transform(x, y)` behaves differently than `fit(x, y).transform(x)` in encoders like LeaveOneOut or CatBoostEncoder.

Care to explain why? I get the same output from `fit_transform(x, y)` and `fit(x, y).transform(x)` , I think I'm missing something here.

> 3. I did not incorporate a special path for binary case because I am leaving it up to the wrapper to deal with string labels.

If I understand correctly, you're making the wrapper to handle string labels? But if so, what does it had to do with the binary case?

janmotl:
>> I am removing the single class before fitting the encoders (i.e. as soon as possible).

> What do you mean?

In the proposal, we first fit _n_ encoders and then remove one of them:
```python
         # train the feature encoders
         for class_name, label in labels.iteritems():
            self.feature_encoders[class_name] = copy.deepcopy(self.feature_encoder).fit(X, label)

         # remove last class to have a more compact encoding
         self.feature_encoders.pop(class_name)
```
Instead of that, we can first remove one class and only then train _n-1_ encoders:
```python
        # drop one label
        labels = labels.iloc[:, 1:] 

        # train the feature encoders
        for class_name, label in labels.iteritems():
            self.feature_encoders[class_name] = copy.deepcopy(self.feature_encoder).fit(X, label)
```
It should be a tiny bit faster this way.


>> It is necessary to respect that fit_transform(x, y) behaves differently than fit(x, y).transform(x) in encoders like LeaveOneOut or CatBoostEncoder.

> Care to explain why? I get the same output from fit_transform(x, y) and fit(x, y).transform(x) , I think I'm missing something here.

That is suspicious. When I run test `test_transform_only_selected(self)` in `category_encoders/tests/test_wrapper.py` in the master branch, I actually assert that the results are different. I am getting different `result1` and `result2` for both, LeaveOneOut and CatBoostEncoder.

A blog [post](https://pkghosh.wordpress.com/2018/06/18/leave-one-out-encoding-for-categorical-feature-variables-on-spark/) about the differences for LeaveOneOut. For training data, we should always use `fit_transform(x, y)`, which uses the longer equation. For testing data, we should always use `transform(x)`, which uses the shorter equation.

>> I did not incorporate a special path for binary case because I am leaving it up to the wrapper to deal with string labels.

> If I understand correctly, you're making the wrapper to handle string labels? But if so, what does it had to do with the binary case?

Currently, the wrapper in the master branch unnecessarily encodes the label even when the label consists only of `{0,1}`. Your proposal takes care of that. On the other hand, when the label consists of `{'a', 'b'}`, the proposal crashes. 

I was feeling too lazy to cover just `{0,1}` scenario, since if you are after speed, then in this special case, you do not need to use the wrapper at all...

PR #221, opened 2019-10-30
Fix OrdinalEncoder with categorical columns containing NaN values

jeffzi:
OrdinalEncoder and OneHotEncoder display incorrect results with categorical columns that include NaN values. 

```python
import numpy as np
import pandas as pd

from category_encoders import OrdinalEncoder, OneHotEncoder

X = pd.DataFrame({
        'A': pd.Categorical(['a', 'b', 'c', 'a', np.nan]),
        'B': ['a', 'b', 'c', 'a', np.nan]
})

# We expect the same results on both columns
OrdinalEncoder().fit_transform(X)
#>                      A  B
0                    1  1
1                    2  2
2                    3  3
3                    1  1
4 -9223372036854775808  4
OneHotEncoder(use_cat_names = True).fit_transform(X)
#>    A_a  A_b  A_c  B_a  B_b  B_c  B_nan
0  1.0  0.0  0.0    1    0    0      0
1  0.0  1.0  0.0    0    1    0      0
2  0.0  0.0  1.0    0    0    1      0
3  1.0  0.0  0.0    1    0    0      0
4  NaN  NaN  NaN    0    0    0      1
```

<sup>Created on 2019-10-30 by the [reprexpy package](https://github.com/crew102/reprexpy)</sup>

The bug results from the fact that pandas doesn't include NaN in categories. This PR fixes the issue in OrdinalEncoder. OneHotEncoder relies on OrdinalEncoder, so it fixes OneHotEncoder as well. 

I also added tests for this specific case.

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -286,7 +286,15 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
             mapping_out = mapping
             for switch in mapping:
                 column = switch.get('col')
-                X[column] = X[column].map(switch['mapping'])
+                col_mapping = switch['mapping']
+                X[column] = X[column].map(col_mapping)
+
+                if util.is_category(X[column].dtype):
+                    if not isinstance(col_mapping, pd.Series):
+                        col_mapping = pd.Series(col_mapping)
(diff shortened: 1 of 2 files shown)


janmotl:
Thank you.

PR #224, opened 2019-11-28
convert index to python list

hershaw:
Since pandas dataframe columns may be pd.Index objects, evaluating them in an if statement may throw an error. converting to a list allows for python to use it's native list boolean evaluation which is pretty robust.

I haven't tested it much at all but there are a fairly few downsides to doing the conversion in the context if an if statement.

Opening the PR just to let you guys know there might be some low hanging fruit available. Will read the guidelines and do a proper one if I find time.


janmotl:
Can you provide an example when the code currently fails?

hershaw:
Updated the description @janmotl with reproduction code and stack trace

janmotl:
Thank you. I applied the fix to all the encoders and added the example as a unit test.

PR #233, opened 2020-01-31
Add KFoldEnconder and unit test

catapulta:
Hi, thanks for the great stuff in this repo. I'd like to contribute with a KFoldEncoder -- it basically performs a K-Fold split and then encodes using the TargetEncoder. For fit_transform, it will return the TargetEncoder output at that fold, for transform, it'll return the average across folds. I also built a simple test that decomposes the above procedure outside the class and compares the outputs. Let me know any comments.

### DIFF
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -22,6 +22,7 @@ from category_encoders.woe import WOEEncoder
 from category_encoders.m_estimate import MEstimateEncoder
 from category_encoders.james_stein import JamesSteinEncoder
 from category_encoders.cat_boost import CatBoostEncoder
+from category_encoders.kfold_encoder import KFoldEncoder
 
 __version__ = '2.1.0'
 
@@ -42,5 +43,6 @@ __all__ = [
     'WOEEncoder',
     'MEstimateEncoder',
     'JamesSteinEncoder',
-    'CatBoostEncoder'
+    'CatBoostEncoder',
+    'KFoldEncoder'
 ]
--- a/category_encoders/kfold_encoder.py
+++ b/category_encoders/kfold_encoder.py
@@ -1,0 +1,220 @@
+"""K Fold encoding"""
+import pandas as pd
+import numpy as np
+from category_encoders.target_encoder import TargetEncoder
+import category_encoders.utils as util
+from sklearn.model_selection import KFold
+from sklearn.base import BaseEstimator, TransformerMixin
+
+__author__ = 'Aldo Marini'
+
+
+class KFoldEncoder(BaseEstimator, TransformerMixin):
+    """K Fold target encoding for categorical features.
+
+    Uses TargetEncoding under the hood, but excludes the current fold's
+    target when calculating the mean target for a level to reduce the effect
+    of outliers.
+
+    Parameters
+    ----------
+
+    K: number of folds.
+    verbose: int
+        integer indicating verbosity of the output. 0 for none.
+    cols: list
+        a list of columns to encode, if None, all string columns will be encoded.
+    drop_invariant: bool
+        boolean for whether or not to drop columns with 0 variance.
+    return_df: bool
+        boolean for whether to return a pandas DataFrame from transform (otherwise it will be a numpy array).
+    handle_unknown: str
+        options are 'error', 'return_nan' and 'value', defaults to 'value', which will impute the target mean.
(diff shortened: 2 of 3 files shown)


janmotl:
Nice. Some quick notes:

1. Do not pass all the settings of KFoldEncoder into the inner encoder. For example, if you pass `drop_invariant=True` into the inner encoder, the inner encoders might return dataframes of different width and KFoldEncoder will fail because it is not prepared for that. Simply pass `drop_invariant=False` into the inner encoder and leave the dropping on the code in KFoldEncoder.
2. Similarly, you must always use `return_df=True` for the inner encoder. Otherwise with `return_df=False` KFoldEncoder will fail.
3. Similarly, you must always use `handle_unknown='value'` for the inner encoder. Otherwise with `handle_unknown='error'` or `handle_unknown='return_nan'` KFoldEncoder may fail.
4. You may copy-paste `get_feature_names` method from TargetEncoder into KFoldEncoder, since you are already setting `self.feature_names`. 
5. Consider converting `X[:] = x / self.K` into `X[:] = x / float(self.K)` to make sure that Python2 always performs floating-point division and not integer division (I did not test the necessity of the cast, but it should not hurt).
6. When not all the string columns are converted (as set by `cols` argument), `x += te.transform(X).values` fails.
7. You may consider generalizing the encoder into a wrapper, which accepts any supervised encoder as an argument (maybe as a Class instead of an instance in order to maintain control over the arguments...). But it is not necessary (better something than nothing).
8. Consider explaining in KFoldEncoder how do you "score" - that you score the whole data with each inner encoder and average their "scores". The reasoning: I was curious about how do you score new data and I had to look into the code in order to answer that. Hence, I assume that other people may want to read how do you do it in the documentation.

If you decide to test KFoldEncoder with `test_encoders` test set, you may want to:
1. Use `K=2` as the default value. Reasoning: Some tests use data sets with less than 5 rows.
2. Use `random_state=2001` or some other constant as the default value. Reasoning: Some tests expect deterministic result.
3. Update `test_get_feature_names_drop_invariant` and `test_get_feature_names` tests to recognize KFoldEncoder as a supervised encoder. 

But of course, consider modifying the tests as you wish.

janmotl:
@Anjum48, do you have some notes?

Anjum48:
I had started writing a wrapper, but I haven't got around to test this in anger yet. I've dropped it into a gist, so you can copy whatever you think is relevant: https://gist.github.com/Anjum48/d958014bbe4cd46a330870fd582db97d

- I agree with @janmotl, a wrapper format would be more convenient
- It would be good to be able to specify the cv method. This might mean adding a `groups` kwarg to make it compatible with all of the sklearn CV methods like `GroupKFold`.

I might be wrong about this, but when using a supervised encoder on test or validation data, you can just use:
```
encoder.fit(X_train, y_train)  # fit the encoder using all of the training data
X_valid_encoded = encoder.transform(X_valid)
```
But for the training data, you have to use something like:
```
X_train_encoded = np.zeros(X.shape)
    for trn, val in kfold.split(X, y):  # Generate out-of-fold encodings to train a model
        encoder.fit(X[trn], y[trn])
        X_train_encoded[val] = encoder.transform(X[val])
```
For that reason, my wrapper only had a `fit_tranform` method, since it will only be used for the training data, and not the validiation/test. I'd be intersted to hear your thoughts on this.

janmotl:
@Anjum48 Good point about the CV method.

For encoding training data with supervised encoders, we should possibly stick with `fit_transform()`, as `README.md` suggests:
> For the transformation of the training data with the supervised methods, you should use fit_transform() method instead of fit().transform(), because these two methods do not have to generate the same result. The difference can be observed with LeaveOneOut encoder, which performs a nested cross-validation for the training data in fit_transform() method (to decrease over-fitting of the downstream model) but uses all the training data for scoring with transform() method (to get as accurate estimates as possible).

After taking that into account, the training loop may look like:
```python
    X_train_encoded = np.zeros(X.shape)
    for trn, val in kfold.split(X, y):  # Generate out-of-fold encodings to train a model
        X_train_encoded[val] = encoder.fit_transform(X[trn], y[trn], X[val])
```

I do not know which approach for data scoring is better. Both approaches look OK.

janmotl:
@Catapulta, I merged the pull request from Anjum48. Fell free to open a new one.

PR #238, opened 2020-03-08
Determine unique values from data, not pandas.cat

### LINKED ISSUE #235: TargetEncoder produces missing values with pandas categorical data
Using the TargetEncoder to transform data with unknown categorical values and `handle_unknown='value'` (default) can produce `NaN`-values. This happens when the input data is a pandas DataFrame with the categorical column of type "category":

```python
import pandas as pd
from category_encoders import TargetEncoder

x = pd.DataFrame(['A','A','B'], dtype='category')
y = pd.Series([0.1,0.1,0.2])
te = TargetEncoder(handle_unknown='value').fit(x.loc[[0, 1]], y.loc[[0, 1]])
te.transform(x.loc[[2]])
>>> NaN
```

When the data is passed in as dtype "object" it works as expected:

```python
import pandas as pd
from category_encoders import TargetEncoder

x = pd.DataFrame(['A','A','B'], dtype='object')  # Notice changed dtype
y = pd.Series([0.1,0.1,0.2])
te = TargetEncoder(handle_unknown='value').fit(x.loc[[0, 1]], y.loc[[0, 1]])
te.transform(x.loc[[2]])
>>> 0.1
```

PGijsbers:
Closes #235.

When using the meta-data of the pandas categorical dtype to determine
all unique values, it might create an encoding for values not present in
the data. Ignoring the meta-data and using the seen unique values
instead ensures proper behavior when a new value occurs in the test
data. It will be handled through the handle_unknown strategy
accordingly.

---

This PR is opened prematurely, sorry for that. I still need to add test the fix and add a unit test.
Unfortunately I encountered issues running the unit tests locally.
I tried to, but got this error while running tests (on Py3.6.10, with pandas 1.0.1 and 0.25.3):

```
Traceback (most recent call last):
  File "~\continuum\anaconda3\envs\ce\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "~\continuum\anaconda3\envs\ce\lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "~\repositories\forks\categorical-encoding\category_encoders\hashing.py", line 203, in require_data
    data_part = self.hashing_trick(X_in=data_part, hashing_method=self.hash_method, N=self.n_components, cols=self.cols)
  File "~\repositories\forks\categorical-encoding\category_encoders\hashing.py", line 391, in hashing_trick
    X_cat.columns = new_cols
  File "~\continuum\anaconda3\envs\ce\lib\site-packages\pandas\core\generic.py", line 5287, in __setattr__
    return object.__setattr__(self, name, value)
  File "pandas\_libs\properties.pyx", line 67, in pandas._libs.properties.AxisProperty.__set__
  File "~\continuum\anaconda3\envs\ce\lib\site-packages\pandas\core\generic.py", line 661, in _set_axis
    self._data.set_axis(axis, labels)
  File "~\continuum\anaconda3\envs\ce\lib\site-packages\pandas\core\internals\managers.py", line 178, in set_axis
    f"Length mismatch: Expected axis has {old_len} elements, new "
ValueError: Length mismatch: Expected axis has 1 elements, new values have 8 elements
```

I am not sure if I am testing with wrong versions here, is there documentation on the developer requirements?
I installed:

```
nose
coverage
unittest2
```

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -320,10 +320,15 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
 
                 nan_identity = np.nan
 
-                # Avoid using pandas category dtype meta-data, see #235
-                categories = X[col].unique().tolist()
-                if util.is_category(X[col].dtype) and X[col].isna().any():
-                    categories += [np.nan]
+                if util.is_category(X[col].dtype):
+                    # Avoid using pandas category dtype meta-data if possible, see #235, #238.
+                    categories = X[col].unique().tolist()
+                    if X[col].dtype.ordered:
+                        categories = [c for c in X[col].dtype.categories if c in categories]
+                    if X[col].isna().any():
+                        categories += [np.nan]
+                else:
+                    categories = X[col].unique()
 
                 index = pd.Series(categories).fillna(nan_identity).unique()
 
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -320,15 +320,13 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
 
                 nan_identity = np.nan
 
+                categories = X[col].unique().tolist()
                 if util.is_category(X[col].dtype):
                     # Avoid using pandas category dtype meta-data if possible, see #235, #238.
-                    categories = X[col].unique().tolist()
                     if X[col].dtype.ordered:
                         categories = [c for c in X[col].dtype.categories if c in categories]
                     if X[col].isna().any():
                         categories += [np.nan]
-                else:
-                    categories = X[col].unique()
 
                 index = pd.Series(categories).fillna(nan_identity).unique()
 
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -320,12 +320,10 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
 
                 nan_identity = np.nan
 
-                if util.is_category(X[col].dtype):
-                    categories = X[col].cat.categories.tolist()
-                    if X[col].isna().any():
-                        categories += [np.nan]
-                else:
-                    categories = X[col].unique()
+                # Avoid using pandas category dtype meta-data, see #235
+                categories = X[col].unique().tolist()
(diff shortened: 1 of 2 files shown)


PGijsbers:
The following old unit test fails:

```python
    def test_pandas_categorical(self):
        X = pd.DataFrame({
            'Str': ['a', 'c', 'c', 'd'],
            'Categorical': pd.Categorical(list('bbea'), categories=['e', 'a', 'b'], ordered=True)
        })

        enc = encoders.OrdinalEncoder()
        out = enc.fit_transform(X)

        th.verify_numeric(out)
        self.assertEqual(3, out['Categorical'][0])
        self.assertEqual(3, out['Categorical'][1])
        self.assertEqual(1, out['Categorical'][2])
        self.assertEqual(2, out['Categorical'][3])
```

It is expecting the mapping to be based on the categorical meta-data.
Following the changes, the output actually follows the mapping `b->1, e->2, a->3`.
I can update the unit test accordingly. But really the exact mapping is not specified.
Should I rewrite the test instead to test that in the transformed categorical column:
 - the first two elements are mapped to the same number, and
 - the third and fourth element are both different from any other element in the output?

janmotl:
If the categorical attribute was defined with `ordered=False`, your proposed test would be right. But if someone bothered to define the order of the categorical values, I think we should honor it. 

Can't we get a list which:

1. honors the ordering in `.cat.categories.tolist()`
2. and contains only values in `.unique().tolist()`

?

PGijsbers:
Yes we can. Follow up question then, when we see an ordered categorical variable defined with categories (in order) `['a' , 'b', 'c']`, and as training data for the OrdinalEncoder only `['a', 'c']`is passed. What mapping do we expect?  `a=0, c=2` or `a=0, c=1`?

janmotl:
I would opt for `a=0, c=1`. Reasoning: If a user wants `a=0, c=2`, it is easy to do it without any library.

PGijsbers:
Done! Let me know if you see room for improvement.

janmotl:
Thank you, @PGijsbers. I have updated `requirements.txt` to make it easier for others to run the tests.

Issue #251, opened 2020-04-30
Count Encoder Documentation missing


pyrrhull:
As far as I can see the Count Encoder Documentation is missing here
http://contrib.scikit-learn.org/category_encoders/.

PR #252, opened 2020-04-30
added count_encoder to docs

makrobios:
Count Encoder is missing from the docs at http://contrib.scikit-learn.org/category_encoders/
see issue [251](https://github.com/scikit-learn-contrib/category_encoders/issues/251#issue-609462519)


wdm0006:
Looks like the new page needs to be added to the toc tree in `index.rst`.

makrobios:
Alright, I added the lines and built the docs locally to confirm.

wdm0006:
Not sure why the tests weren't passing on this, could you push something new to your fork to trigger the build to run again and see if it was a temporary thing? I think we were still messing around with the CI/CD at the time.

makrobios:
Tried out as you suggested but didn't pass the tests again. It seems CountEncoder is also missing from `"__all__"` in `category_encoders/__init__.py` which I fixed in the PR. 
That is why "test_classification"  in `category_encoders/tests/test_encoders.py `is skipped in master for CountEncoder but fails here. 
Any suggestion where to go from here? I'd be happy to help and would appreciate some guidance.

makrobios:
Just reverted above mentioned changes in `__init__.py` and as expected, all tests passed. Not sure if this is correct though.

wdm0006:
Hmmm yeah, looks like your reversion is just leading to test_encoders to skip the CountEncoder, which fails if it's not skipped.  The correct thing would be to have it in the __all__ and resolve the test failures in the actual implementation (I understand that's not really the content of your PR here).

I'm not personally super familiar with the count encoder implementation, so may need to carve out some time to take a closer look into why the tests are actually failing.

makrobios:
I see, i started digging through the tests and there are 11 failing, and they are all in `tests/test_encoders.py`. Do you think it is better to open another issue or modify this one?

makrobios:
see  #260

Issue #254, opened 2020-05-05
Under certain conditions, the index is reset after being processed by multi-process HashingEncoder


pesuchin:
## Expected Behavior

The index should not be changed after being processed by HashingEncoder in either single or multi-process.

## Actual Behavior

Under certain conditions, the index is changed after being processed by multi-process HashingEncoder.

## Steps to Reproduce the Problem

  1. Prepare a DataFrame.
  1. Cut out some rows to make the index partially holey.
  1. Assign HashingEncoder's col to process only some of the columns.
  1. Fit and transform HashingEncoder.

### Concrete example: Single process HashingEncoder

In the case of a single process, the index will not be changed as follows.

```
import category_encoders as encoders
import pandas as pd
from pandas.testing import assert_index_equal

columns = ['column1', 'column2', 'column3', 'column4']
df = pd.DataFrame([[i, i, i, i] for i in range(10)], columns=columns)
df = df.iloc[2:8, :]
target_columns = ['column1', 'column2', 'column3']

single_process_encoder = encoders.HashingEncoder(max_process=1, cols=target_columns)
single_process_encoder.fit(df, None)
df_encoded_single_process = single_process_encoder.transform(df)

print('origin:', list(df.index))
print('after HashingEncoding: ', list(df_encoded_single_process.index))
print(df.index.equals(df_encoded_single_process.index))
-----------------------[output]-----------------------
origin: [2, 3, 4, 5, 6, 7]
after HashingEncoding:  [2, 3, 4, 5, 6, 7]
True
```

### Concrete example: Multi process HashingEncoder

In the case of multi-process, the index will be changed as follows.

```
import category_encoders as encoders
import pandas as pd
from pandas.testing import assert_index_equal

columns = ['column1', 'column2', 'column3', 'column4']
df = pd.DataFrame([[i, i, i, i] for i in range(10)], columns=columns)
df = df.iloc[2:8, :]
target_columns = ['column1', 'column2', 'column3']

multi_process_encoder = encoders.HashingEncoder(max_process=0, cols=target_columns)
multi_process_encoder.fit(df, None)
df_encoded_multi_process = multi_process_encoder.transform(df)
print('origin: ', list(df.index))
print('after HashingEncoding: ', list(df_encoded_multi_process.index))
print(df.index.equals(df_encoded_multi_process.index))
-----------------------[output]-----------------------
origin:  [2, 3, 4, 5, 6, 7]
[...]

PR #255, opened 2020-05-05
Fix: #254 avoid resetting index in multi-process HashingEncoder

pesuchin:
Fix: #254 

## Proposed Changes

  - Disable ignore_index option
    - From what I see, it doesn't seem necessary to specify ignore_index, so I suggest not to do so.
  - add test code


wdm0006:
Looks good, thanks.

PR #260, opened 2020-07-10
Count encoder

makrobios:
## Fixes 

- test_metamorphic
- test handle_missing
- test_classification
- test_get_feature_names

## Proposed Changes

- added count_encoder to docs
- Updated docstrings for CountEncoder
- renamed handle_missing default option from 'count' to 'value'
- added 'value' option to handle_unknown
- added method "get_feature_names" to CountEncoder

~~## TODO:~~
~~tests that were skipped previously and failed and still fail in test_encoders.py::TestEncoders are:~~
~~- handle_unknown test_handle_unknown_value~~
~~- combine_min_nan_groups makes test_column_transformer and test_sklearn_compliance fail~~


wdm0006:
Looks good, thanks for taking on a bigger chunk of work than originally intended

janmotl:
Note that the following test:
```python
    def test_none_in_omitted(self):
        enc = encoders.CountEncoder(cols=['A'])
        X = pd.DataFrame({'A':['a','b',None], 'B':['c','d',None]})

        out = enc.fit_transform(X)

        self.assertTrue(out['B'][2] is None)
```
would not pass anymore because all "missing values" are now replaced with `np.nan`.

Originally, there was:
```python
X.loc[:, self.cols] = X.fillna(value=np.nan)
```
But now there is:
```python
X.fillna(value=np.nan, inplace=True)
```

I believe we should not alter ignored columns, unless there is a compelling reason to do, as there can be a reason, why the ignored columns use the specific "missing value" representation.

If the described behavior is undesired, consider generalizing the test to all encoders (placing it in `test_encoders`).

makrobios:
I think you make a valid point there. The problem with the original code is that it throws the exception
`
ValueError: Cannot setitem on a Categorical with a new category, set the categories first
`
 in 
`test_encoders.py::TestEncoders::test_metamorphic` (line 614).

makrobios:
Here might be a possible solution for the problem #261

PR #261, opened 2020-07-20
edit _transform_count_encode

makrobios:
moved X.fillna... inside for loop so that it handles only columns
present in self.cols individually and not the whole dataframe
#260 

  -
  -
  -

### DIFF
--- a/category_encoders/count.py
+++ b/category_encoders/count.py
@@ -271,9 +271,11 @@ class CountEncoder(BaseEstimator, TransformerMixin):
     def _transform_count_encode(self, X_in, y):
         """Perform the transform count encoding."""
         X = X_in.copy(deep=True)
-        X.fillna(value=np.nan, inplace=True)
 
         for col in self.cols:
+
+            X[col] = X.fillna(value=np.nan)[col]
+
             if self._min_group_size is not None:
                 if col in self._min_group_categories.keys():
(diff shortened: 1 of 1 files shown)


janmotl:
Looks good to me. Consider adding a unit test to make sure that the issue does not return (personally, I would have preffered a parametrized test in test_encoders to cover all the encoders).

makrobios:
I am considering the tests but don't know how to proceed. What should the tests be against and which parameters to use? Isn't test_metamorphic already doing the job here?

janmotl:
Metamorphic tests encoded features. But we also have to test non-encoded features. The test to add:

```python
    def test_ignored_columns_are_untouched(self):
        # Make sure None values in ignored columns are preserved.
        # See: https://github.com/scikit-learn-contrib/category_encoders/pull/261
        X = pd.DataFrame({'col1': ['A', 'B', None], 'col2': ['C', 'D', None]})
        y = [1, 0, 1]

        for encoder_name in (set(encoders.__all__)):
                enc = getattr(encoders, encoder_name)(cols=['col1'])
                out = enc.fit_transform(X, y)
                self.assertTrue(out.col2[2] is None)
```

makrobios:
Great, so I added your test to test_encoders.py.

PR #266, opened 2020-08-13
is_categorical deprecated

philip-khor:
Getting FutureWarning that is_categorical is deprecated. See pandas-dev/pandas#33385

## Proposed Changes

  - Change `is_categorical()` to `is_categorical_dtype()` 
  -
  -

### DIFF
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -18,7 +18,7 @@ def convert_cols_to_list(cols):
         return list(cols)
     elif isinstance(cols, tuple):
         return list(cols)
-    elif pd.api.types.is_categorical(cols):
+    elif pd.api.types.is_categorical_dtype(cols):
         return cols.astype(object).tolist()
 
     return cols


PGijsbers:
I don't mean to intrude, but I encountered this issue myself and would love to see this updated in the package :)

One word of caution though, simply replacing the call does break for users with pandas `1.0.x` or below.
In case it is desired to keep support for `pandas<1.1.0`, call the appropriate method depending on the pandas version.
The cleanest way is probably to just check if they have `is_categorical_dtype` and fall back on `is_categorical` otherwise.

MatteoFelici:
Hi! Is there an update on this?

RoyalTS:
Is there any chance of this getting merged and of the package getting updated anytime soon? Would be happy to do the work if more work needs doing.

jonathanng:
```import warnings

ignores = [
    ('category_encoders', FutureWarning, 'is_categorical is deprecated and will be removed in a future version.  Use is_categorical_dtype instead')
    # add addition warnings you want to ignore
]

for module, category, message in ignores:
    warnings.filterwarnings('ignore', module=module, category=category, message=message)
```

PR #269, opened 2020-09-10
support k fold in TargetEncoder

### LINKED ISSUE #133: K-fold target encoding to mitigate overfitting risks
Dear maintainers, would you please consider adding into TargetEncoder module ability to compute target means by category in out-of-fold fashion using custom folds generator?

That way, at a fitting stage we would not just compute single values of smoothing for given category of a given column using all available rows. Instead, we would iterate y array fold by fold, computing mean/smoothing over each **train** fold and using it for rows of each corresponding **test** fold.
This could really allow to reduce overfitting by introducing variability while still conforming to the peculiarities of underlying data.

Coding-wise, I think it could be implementing by modifying just one procedure, namely **target_encode** in **target_encoder.py**: tmp[val]['smoothing'] would become a list instead of a scalar value, and we would need adding 2 loops of such kind:

```python
for nfold,(train_idx, valid_idx) in enumerate(folding.split(X, y)):
    #treat X[train_idx] and y[train_idx] as we are treating X and y now 
    #while computing tmp[val]['smoothing']

    #only use tmp[val]['smoothing'].append(cust_smoothing) to store results as it will be a list, not scalar
```

and 

```python
for nfold,(train_idx, valid_idx) in enumerate(folding.split(X)):
    transformed_column.loc[X.loc[valid_idx,column] == val] = switch.get('mapping')[val]['smoothing'][nfold]
    #to apply learned values at a transform step
```
**folding=KFold(1)** would then become an additional parameter of TargetEncoder object's signature, default param of it behaving exactly like it does now.

Do you think this is feasible to implement?

guolinke:
Fixes #133 

## Proposed Changes

  - add `nfolds` and `stratified` to `TargetEncoder`

### DIFF
--- a/category_encoders/target_encoder.py
+++ b/category_encoders/target_encoder.py
@@ -85,7 +85,7 @@ class TargetEncoder(BaseEstimator, util.TransformerWithTargetMixin):
     """
 
     def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, handle_missing='value',
-                     handle_unknown='value', min_samples_leaf=1, smoothing=1.0, nfolds=4, stratified=False):
+                     handle_unknown='value', min_samples_leaf=1, smoothing=1.0, nfolds=1, stratified=False):
         self.return_df = return_df
         self.drop_invariant = drop_invariant
         self.drop_cols = []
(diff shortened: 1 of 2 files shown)


category_encoders/target_encoder.py
```diff
@@ -143,12 +153,13 @@ def fit(self, X, y, **kwargs):
...
         
-        X_temp = self.transform(X, override_return_df=True)
+        X_temp = self.transform(X, y, override_return_df=True)
         self.feature_names = list(X_temp.columns)
 
         if self.drop_invariant:
             self.drop_cols = []
-            X_temp = self.transform(X)
+            # fixme: why here call again?
+            # X_temp = self.transform(X, y)
```
guolinke, line 162: I don't know why here call transform again.

guolinke:
just find this could be achieved by `NestedCVWrapper(TargetEncoder())`. closed.

Issue #71, opened 2018-05-25
Is frequency encoding covered in this?


sebastiangonsal:
Is frequency encoding covered with any of these encoders ?

wdm0006:
Nope, do you have a reference for it?

ornelaxh:
@sebastiangonsal do you mean simply the counts of every category?

MatteoFelici:
If this is still needed, I can work on that.

MatteoFelici:
Isn't this the `CountEncoder`? If so, I suggest closing this issue.

PR #271, opened 2020-09-11
add on-the-fly frequency encoder

guolinke:
Fixes and improves #71

This is like the combination of count and CatBoost. The frequency is calculated on-the-fly.
This is used for the chronological data.


guolinke:
Although this is an unsupervised encoding, but it uses different encodings for the training set and test set. So the label is used for distinguishing the training set.

Issue #264, opened 2020-08-03
Some confusions about BinaryEncoder


AshiakerWang:
It seems that there are some **redundant bits** in the output of [BinaryEncoder](http://contrib.scikit-learn.org/category_encoders/binary.html).

For example, in the following code :

```
import pandas as pd
from IPython.core.display import display
from category_encoders import *
from sklearn.datasets import load_boston

bunch = load_boston()
X = pd.DataFrame(bunch.data, columns=bunch.feature_names)
enc = BinaryEncoder(cols=['RAD']).fit(X)
numeric_dataset = enc.transform(X)
display(numeric_dataset.head(1000))  
```
The `RAD_0`  column of the output `numeric_dataset ` contains only zeros:
```
>>>numeric_dataset['RAD_0'].value_counts()
0    506
```
Number of  **log2⌈m⌉ bits**  is sufficient to encode a column which contains m possible categories, isn't it ? 
In the example above,  log2⌈9⌉ bits = 4 is sufficient to encode the `'RAD'` column, where 9 is the number of categories in the `'RAD'` column.
Therefore, I suppose the `RAD_0`  column of the output is redundant.
It will be appreciated if you could give me some suggestions about the rational of introducing such extra column (i.e. `RAD_0`) in the output.

kerrickstaley:
I opened a PR to fix this issue: https://github.com/scikit-learn-contrib/category_encoders/pull/296

PR #296, opened 2021-04-12
Fix BaseNEncoder number of output columns

kerrickstaley:
BaseNEncoder encoder used an incorrect formula for calculating the number of required bits in the output. If there are `nvals` distinct values and we reserve one encoding to represent "missing or unknown", then the correct number of bits is `ceil(log(nvals + 1, base))`. However, the code was previously using the formula `ceil(log(nvals, base)) + 1`.

Fixes https://github.com/scikit-learn-contrib/category_encoders/issues/264

## Proposed Changes
  - Change the formula to `ceil(log(nvals + 1, base))`.
  - Switch the formula to use integer math so we don't have to worry about floating point rounding errors.
  - Add a test.
  - Fix a non-deterministic test.

### DIFF
--- a/category_encoders/basen.py
+++ b/category_encoders/basen.py
@@ -12,6 +12,27 @@ import warnings
 __author__ = 'willmcginnis'
 
 
+def _ceillogint(n, base):
+    """
+    Returns ceil(log(n, base)) for integers n and base.
+
+    Uses integer math, so the result is not subject to floating point rounding errors.
+
+    base must be >= 2 and n must be >= 1.
+    """
+    if base < 2:
+        raise ValueError('base must be >= 2')
+    if n < 1:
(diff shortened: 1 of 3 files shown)


kerrickstaley:
@janmotl @wdm0006 this is ready review, could you take a look?

kerrickstaley:
I updated the code to use the formula `ceil(log(nvals + 1, base))`, which has a more obvious derivation than `floor(log(nvals, base)) + 1`. The two formulas return the same results for integers `nvals` and `base`.

wdm0006:
Looks like the basen encoding tests are passing fine but theres an issue in master for the GLM encoding unrelated to this PR. Going to go ahead and merge this, thanks for the patience.

Issue #302, opened 2021-05-31
[new features]: Quantile Encoder


cmougan:
Implementation of Quantile Encoder from the publication (https://arxiv.org/abs/2105.13783)

Quantile Encoder: Tackling High Cardinality Categorical Features in Regression Problems
Carlos Mougan, David Masip, Jordi Nin, Oriol Pujol

Zainny1234:
Hi cmougan.Thanks for sharing the paper.I am trying to create a rental price avm model, but have categorical values with high cardinality. I am going through the paper, having slight difficulty in grasping the methodology.Is there a coded solution anywhere for this?

cmougan:
Hi @Zainny1234! The usage of this encoder follows the same structure than the rest of category_encoders packages.
```python
    >>> from category_encoders import *
    >>> import pandas as pd
    >>> from sklearn.datasets import load_boston
    >>> bunch = load_boston()
    >>> y = bunch.target
    >>> X = pd.DataFrame(bunch.data, columns=bunch.feature_names)
    >>> enc = QuantileEncoder(cols=['CHAS', 'RAD']).fit(X, y)
    >>> numeric_dataset = enc.transform(X)
    >>> print(numeric_dataset.info())
    <class 'pandas.core.frame.DataFrame'>
    RangeIndex: 506 entries, 0 to 505
    Data columns (total 13 columns):
    CRIM       506 non-null float64
    ZN         506 non-null float64
    INDUS      506 non-null float64
    CHAS       506 non-null float64
    NOX        506 non-null float64
    RM         506 non-null float64
    AGE        506 non-null float64
    DIS        506 non-null float64
    RAD        506 non-null float64
    TAX        506 non-null float64
    PTRATIO    506 non-null float64
    B          506 non-null float64
    LSTAT      506 non-null float64
    dtypes: float64(13)
    memory usage: 51.5 KB
    None
```

While the PR does not get accepted, you can use the package in
```python
from sktools import QuantileEncoder
```

PR #303, opened 2021-05-31
Quantile encoder

cmougan:
This PR (#302), implements two methods from a recently published paper at a conference (MDAI 2021).

> [Quantile Encoder: Tackling High Cardinality Categorical Features in Regression Problems (Carlos Mougan, David Masip, Jordi Nin, Oriol Pujol) ](https://arxiv.org/abs/2105.13783)

Encoding methods, full technical development can be followed in the paper:
- Quantile Encoder
- Summary Encoder

Tests are implemented and passed.
Scikit learn API semantics
Docs is extended

If I missed something or any comments, please let me know :)

### DIFF
--- a/category_encoders/quantile_encoder.py
+++ b/category_encoders/quantile_encoder.py
@@ -1,0 +1,334 @@
+"""Target Encoder"""
+__author__ = "david26694", "cmougan"
+
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator
+from category_encoders.ordinal import OrdinalEncoder
+from category_encoders.m_estimate import MEstimateEncoder
+import category_encoders.utils as util
+from sklearn.utils.random import check_random_state
+class QuantileEncoder(BaseEstimator, util.TransformerWithTargetMixin):
+    """Quantile Encoding for categorical features.
+    This a statistically modified version of target MEstimate encoder where selected features
+    are replaced the statistical quantile instead than the mean. Replacing with the
+    median is a particular case where self.quantile = 0.5. In comparison to MEstimateEncoder
+    it has two tunable parameter `m` and `quantile`
+    Parameters
+    ----------
+    verbose: int
+        integer indicating verbosity of the output. 0 for none.
+    quantile: int
+        integer indicating statistical quantile. ´0.5´ for median.
+    m: int
+        integer indicating the smoothing parameter. 0 for no smoothing.
+    cols: list
+        a list of columns to encode, if None, all string columns will be encoded.
+    drop_invariant: bool
+        boolean for whether or not to drop columns with 0 variance.
+    return_df: bool
+        boolean for whether to return a pandas DataFrame from transform (otherwise it will be a numpy array).
+    handle_missing: str
+        options are 'error', 'return_nan'  and 'value', defaults to 'value', which returns the target quantile.
+    handle_unknown: str
+        options are 'error', 'return_nan' and 'value', defaults to 'value', which returns the target quantile.
+    Example
+    -------
+    >>> from category_encoders import *
+    >>> import pandas as pd
+    >>> from sklearn.datasets import load_boston
+    >>> bunch = load_boston()
+    >>> y = bunch.target
+    >>> X = pd.DataFrame(bunch.data, columns=bunch.feature_names)
+    >>> enc = QuantileEncoder(cols=['CHAS', 'RAD']).fit(X, y)
+    >>> numeric_dataset = enc.transform(X)
+    >>> print(numeric_dataset.info())
+    <class 'pandas.core.frame.DataFrame'>
+    RangeIndex: 506 entries, 0 to 505
+    Data columns (total 13 columns):
+    CRIM       506 non-null float64
+    ZN         506 non-null float64
+    INDUS      506 non-null float64
+    CHAS       506 non-null float64
+    NOX        506 non-null float64
+    RM         506 non-null float64
+    AGE        506 non-null float64
+    DIS        506 non-null float64
+    RAD        506 non-null float64
+    TAX        506 non-null float64
+    PTRATIO    506 non-null float64
+    B          506 non-null float64
+    LSTAT      506 non-null float64
+    dtypes: float64(13)
+    memory usage: 51.5 KB
+    None
+    References
+    ----------
+    .. [1] A Preprocessing Scheme for High-Cardinality Categorical Attributes in Classification and Prediction Problems, equation 7, from https://dl.acm.org/citation.cfm?id=507538
+    .. [2] On estimating probabilities in tree pruning, equation 1, from https://link.springer.com/chapter/10.1007/BFb0017010
+    .. [3] Additive smoothing, from https://en.wikipedia.org/wiki/Additive_smoothing#Generalized_to_the_case_of_known_incidence_rates
+    .. [4] Target encoding done the right way https://maxhalford.github.io/blog/target-encoding/
+    """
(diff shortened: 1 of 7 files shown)


cmougan:
@janmotl @wdm0006

cmougan:
Hi @PaulWestenthanner 

Many thanks for your answer.

We will be very happy to contribute to your  package. It makes all the sense to have it as another method in category_encoders.

So, yes please, let's merge.

The only reason  to keep it alive in stools it's because  the congress proceedings  paper points there. Still we will like to merge.

Let us know what  we can do.

category_encoders/quantile_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+__author__ = "david26694", "cmougan"
+
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator
+from category_encoders.ordinal import OrdinalEncoder
+from category_encoders.m_estimate import MEstimateEncoder
+import category_encoders.utils as util
+from sklearn.utils.random import check_random_state
+class QuantileEncoder(BaseEstimator, util.TransformerWithTargetMixin):
```
PaulWestenthanner, line 11: Please add 2 blank lines (PEP-8) before the class definition. Also consider just formatting using black (which you actually did right? maybe that got through somehow- https://github.com/psf/black).

category_encoders/quantile_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+    """Quantile Encoding for categorical features.
+    This a statistically modified version of target MEstimate encoder where selected features
+    are replaced the statistical quantile instead than the mean. Replacing with the
+    median is a particular case where self.quantile = 0.5. In comparison to MEstimateEncoder
+    it has two tunable parameter `m` and `quantile`
+    Parameters
+    ----------
+    verbose: int
+        integer indicating verbosity of the output. 0 for none.
+    quantile: int
```
PaulWestenthanner, line 21: not an integer right? It's a float between 0 and 1.

category_encoders/quantile_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+        options are 'error', 'return_nan' and 'value', defaults to 'value', which returns the target quantile.
+    Example
+    -------
+    >>> from category_encoders import *
+    >>> import pandas as pd
+    >>> from sklearn.datasets import load_boston
+    >>> bunch = load_boston()
+    >>> y = bunch.target
+    >>> X = pd.DataFrame(bunch.data, columns=bunch.feature_names)
+    >>> enc = QuantileEncoder(cols=['CHAS', 'RAD']).fit(X, y)
```
PaulWestenthanner, line 43: as a user i think it would be nice to have the tunable parameters mentioned here like 
```enc = QuantileEncoder(cols=['CHAS', 'RAD'], quantile=0.5, m=1.0).fit(X, y)```

File: category_encoders/quantile_encoder.py [hunk 4]
```diff
@@ -0,0 +1,291 @@
...
+    median is a particular case where self.quantile = 0.5. In comparison to MEstimateEncoder
+    it has two tunable parameter `m` and `quantile`
+    Parameters
+    ----------
+    verbose: int
+        integer indicating verbosity of the output. 0 for none.
+    quantile: int
+        integer indicating statistical quantile. ´0.5´ for median.
+    m: int
+        integer indicating the smoothing parameter. 0 for no smoothing.
```
PaulWestenthanner, line 24: this is (like in m-estimator) a float, right?

category_encoders/quantile_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+                    )
+
+        return self
+
+    def fit_quantile_encoding(self, X, y):
+        mapping = {}
+
+        # Calculate global statistics
+        prior = self._quantile = np.quantile(y, self.quantile)
+        self._sum = y.sum()
```
PaulWestenthanner, line 173: you actually don't use neither `self._sum` nor `self._count`, do you? So I think we can just drop these two lines. Also the `self._quantile` does not need to be defined

category_encoders/quantile_encoder.py
```diff
@@ -0,0 +1,291 @@
...
+        Returns
+        -------
+        feature_names: list
+            A list with all feature names transformed or added.
+            Note: potentially dropped features are not included!
+        """
+
+        if not isinstance(self.feature_names, list):
+            raise ValueError(
+                "Must fit data first. Affected feature names are not known " "before."
```
PaulWestenthanner, line 288: Is there a reason not to write `"... known before"` instead of `"... known " "before"`

docs/source/index.rst
```diff
@@ -85,6 +86,8 @@ Contents:
    targetencoder
    woe
    wrapper
+   quantile
+   summary
```
PaulWestenthanner, line 90: the summary encoder is only in sktools and not here

PaulWestenthanner:
Hi @cmougan 
I just reviewed your code and remarked some minor changes - many of them rather cosmetic. Overall the code is quite nice and will be a good improvement to the library.   
Apart from the comments I'd like to point out that I'm not quite happy with to copy-paste of some functionality (e.g. dropping low-variance columns). However I think it's okay here to just copy-paste since most other encoders did the same and we'll need a separate issue to clean it up.  
Regarding the `SummaryEncoder`: if you want to you can move it over here as well.

category_encoders/quantile_encoder.py
cmougan: Would this be the formatting?
 `python -m black category_encoders/quantile_encoder.py`

cmougan:
@david26694

cmougan:
Many thanks @PaulWestenthanner  for the detailed review.

I just pushed the Summary Encoder. If I have not missed anything, everything should be there.

tests/test_quantile_encoder.py
```diff
@@ -65,6 +61,58 @@ def test_new_category(self):
...
+            quantile_results[self.col].values,
+            summary_results[f"{self.col}_{percentile}"].values,
         )
+
+    def test_several_quantiles(self):
+
+        for quantile in [0.1, 0.5, 0.9]:
+            self.assert_same_quantile(quantile)
+
+    def test_several_quantiles(self):
```
PaulWestenthanner, line 99: this test has the same name as the above one. Is the above one really necessary? If so we should probably rename them

cmougan:
Hi @PaulWestenthanner I believe we just covered your comments. We also did a review of possible things that we might have left out.

Do you see anything else that might be worth to improve? Or that we missed out?

Many thanks for your suggestions,  we really appreciate them :)

PaulWestenthanner:
Hi @cmougan @david26694 

Thanks for adding the tests. This was a good catch!
I took the time to fix the tests. This was mainly because the iterative structure did not comply with other encoders.
Please find my code here: https://github.com/PaulWestenthanner/categorical-encoding/tree/quantileEncoder
It's a little mix of option 1 and 2 since a little code is copied. However, the part of the code that pre-processes X and y and checks for nulls and so on is copied for more or less any encoder in the repo. This issue should be tackled at some point but not now.

I'd suggest we merge my branch into yours and then into master?

cmougan:
Hi @PaulWestenthanner @david26694 

Waw, impresive coding.

Just merged your branch.

cmougan:
@PaulWestenthanner and now?

PaulWestenthanner:
Great! Thanks for your effort. LGTM & Merge

PR #304, opened 2021-06-01
Fix bad WOE scores.

slundberg:
I found that the WOEEncoder was not giving the right scores, and it was because of the the agg stats not being correct from pandas. The stats did work if we grouped by a numpy array instead of a pandas col though. So that's the fix offered here.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/category_encoders/woe.py
+++ b/category_encoders/woe.py
@@ -253,7 +253,7 @@ class WOEEncoder(BaseEstimator, util.TransformerWithTargetMixin):
             col = switch.get('col')
             values = switch.get('mapping')
             # Calculate sum and count of the target for each unique value in the feature col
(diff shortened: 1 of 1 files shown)


PaulWestenthanner:
Hi @slundberg 
could you please provide an example of the error? The example given in the doctest seems to work.  
I'm a little concerned to change this here since for example the M-Estimate (https://github.com/scikit-learn-contrib/category_encoders/blob/master/category_encoders/m_estimate.py#L259) uses a very similar line. Is this broken/incorrect in your case as well?

PaulWestenthanner:
This issue should be fixed by aligning the indices as merged in #320 
@slundberg Could you please confirm?

PR #306, opened 2021-06-22
single row hashing encoder transform error

jonathanglima:
When using the transform for a small dataset the hashing encoder hangs because self.max_sample == 0. This becomes a problem when slicing the dataset and the multiprocessing part of hashing encoder hangs indefinitely. The test shows this problem and this simple if fixes it.

First PR, please forgive me if I'm out of guidelines. I've reviewed the guidelines file, but anyway I might have missed something


jonathanglima:
I've ran the tests and only glmm is having errors, but this was already broken before my changes

jonathanglima:
Any news on this? I got stuck on this error for the second time

Issue #317, opened 2021-10-10
Replace broken link in LeaveOneOutEncoder docs


PaulWestenthanner:
## Expected Behavior
Working link to further readings on how the leave-on-out-encoder works

## Actual Behavior
404 Error on kaggle page. It seems that the whole competition was removed.

## Steps to Reproduce the Problem

Klick link.

## Solutions

Use either
https://pkghosh.wordpress.com/2018/06/18/leave-one-out-encoding-for-categorical-feature-variables-on-spark/
or 
https://datascience.stackexchange.com/questions/10839/what-is-difference-between-one-hot-encoding-and-leave-one-out-encoding

Unfortunately a paper cannot be found.

Issue #237, opened 2020-03-06
CountEncoder not working?


JPFrancoia:
Hi,

I'm having trouble getting the `CountEncoder` working. Calling `fit_transform` on an array-like structure doesn't seem to "do anything". Moreover the output seems slightly different compared to the output of a sklearn encoder. I think the API isn't fully compatible with sklearn, or it's a bug. Here is a snippet to reproduce the issue:

```python
import category_encoders as ce
import pandas as pd
from sklearn.datasets import load_boston

# prepare some data
bunch = load_boston()
X = pd.DataFrame(bunch.data, columns=bunch.feature_names)

enc = ce.CountEncoder()

enc.fit(X["RAD"])

output = enc.transform(X["RAD"])
```

Input (`X["RAD"]` before any transformation):

```
0      1.0
1      2.0
2      2.0
3      3.0
4      3.0
      ... 
501    1.0
502    1.0
503    1.0
504    1.0
505    1.0
Name: RAD, Length: 506, dtype: float64
```

Output:
```
     RAD
0    1.0
1    2.0
2    2.0
3    3.0
4    3.0
..   ...
501  1.0
502  1.0
503  1.0
504  1.0
505  1.0
```

Basically the output isn't what I expected. If I use a "normal" `LabelEncoder` from sklearn, here is what I get:

```
In [5]: le = LabelEncoder()                                                                                                                                                                   

In [6]: le.fit_transform(X["RAD"])                                                                                                                                                            
Out[6]: 
array([0, 1, 1, 2, 2, 2, 4, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 3, 3, 3, 3, 3,
       3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 2, 2, 2, 2, 2,
       2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 2, 4, 1, 4, 7, 7, 7, 7, 7, 7, 2, 3,
       3, 3, 3, 3, 3, 3, 3, 3, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 2, 2, 2, 2,
       1, 1, 1, 1, 3, 3, 3, 1, 1, 1, 1, 1, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4,
[...]

Issue #166, opened 2019-01-05
Metadata for encoders


janmotl:
It should be possible to programmatically differentiate between encoders that:
1. Do not require any target during fitting (like OneHotEncoder).
2. Require some target during fitting (like TargetEncoder).
3. Require binary target during fitting (like WOE).

This information would be useful for:
1. Parameterized tests.
2. Users that wonder, which encoder they may use.

Proposed implementation:
Like in scikit-learn.

amueller:
How would you do that with scikit-learn / how do you want to implement this?

janmotl:
I though that there is some standardized way. But I did not find any reference. Options:
1. Create a few abstract classes. And based on the inheritance we could learn the properties of the encoders.
2. The encoders could implement a couple of methods like 'accepts_continuous_target()' and 'accepts_binary_target()' or implement a single method like [supportsCapability](https://github.com/rapidminer/rapidminer-studio/blob/3a2da3e40bbadaf5515e04093867798c27d0654d/src/main/java/com/rapidminer/operator/learner/bayes/NaiveBayes.java) in RapidMiner.
3. The encoders could have attributes like 'accepts_continuous_target' and 'accepts_binary_target'.

amueller:
The easy / hackish thing to do would be inspecting the signature and see if y is a required argument.
There's hopefully soon estimator tags in sklearn which will allow you to specify this kind of information but they are not really standardized yet.

janmotl:
I found it: https://scikit-learn.org/stable/developers/contributing.html#estimator-types

amueller:
Yes, but that doesn't tell you for a transformation whether it requires ``y`` to fit (I probably wrote that section).

janmotl:
In that case it looks like you are the right person to discuss it.

I am not picky about the used names or the exact mechanism in which it is implemented. The required functionality could be implemented with the following tags:
 - [ ] supports continuous target
 - [ ] supports binomial target

I used term "target" instead of the more common "label", because "target" works well for both, continuous and discrete vectors, while "label" is arguably appropriate only for discrete vectors.

While some of the encoders require the target to take values {0,1}, I believe that they should be refactored in some distant future to support targets like {'no', 'yes'}, {'negative', 'positive'} or any other set of exactly two values. Hence, I prefer the more general and future-proof term "binomial" instead of "binary", which would suggest that the target takes only values {0,1}.

[...]

Issue #122, opened 2018-09-08
All encoders should be recyclable


janmotl:
Currently, when we rerun `fit()` on a trained encoder, the encoder does not get retrained. 

The only notable exception is Leave One Out encoder (thanks to work of @JohnnyC08):
```python    
import unittest

def test_recyclability(self):
     for encoder_name in encoders.__all__:
         with self.subTest(encoder_name=encoder_name):
             # when we use the same encoder on two different datasets, it should not explode
             X_a = pd.DataFrame(data=['1', '2', '2', '2', '2', '2'], columns=['col_a'])
             X_b = pd.DataFrame(data=['1', '1', '1', '2', '2', '2'], columns=['col_b']) # different values and name
             y_dummy = [True, False, True, False, True, False]
             enc = getattr(encoders, encoder_name)()
             enc.fit(X_a, y_dummy)
             enc.fit(X_b, y_dummy)
             verify_numeric(enc.transform(X_b))
```

PR #325, opened 2021-11-28
Refactor/base class

PaulWestenthanner:
### Streamline Encoders
By introducing a `BaseEncoder` and a `Un/SupervisedTransformerMixin` that almost (c.f. below) all encoders inherit from the code is much more streamlined. This removes the boilerplate when implementing new encoders and hence makes it easier for new contributor.  
 
### Following DRY Principle
Common components are reused:
  - When fitting: input conversion, selecting cols, `handle_missing='error'`, setting `self.feature_names`, determine and handle invariant columns
  - When transforming: input conversion, checking if fitted, checking correct input dimension, `handle_missing='error'`
  - `get_feature_names` function

### No Breaking Changes
Note that the API does not change. At least I did not have to change a single test and all our tested cases still work. Since the library is rather well tested, there is reason to be optimistic that such a big change does not break anything.

### Minor Changes
  - Rename `drop_cols` to `invariant_cols` since it is more precise to state the (only) reason why these columns are dropped.
  - Apply refit with different columns from LOO encoder to all others.
  - Introduce a metadata flag if an ordinal pre-encoder is fitted. This might be useful for further abstraction.
  - Started to add type annotations in some places
  - Streamline OneHotEncoder to return the input without converting if no categorical columns are present. All other encoders have this behaviour. Probably setting `return_df = False` and passing an `numpy.ndarray` would even throw an error in the current implementation.

### Major
#166 At least a big step towards that goal. @janmotl suggests adding metadata to encoders via base classes. I'm introducing the base classes here, but only with a minimum of metadata. More can be added very easily.

### Minor
#317  Broken link
#237 Warning if transform does not do anything (see discussion)
#122 This is also partially addressed. The fix for LOO-Encoder is deployed to all other encoders. However this only ensures that the `cols` parameter is treated correctly when calling fit twice. Not the other params. The solution would be probably to define a fitting-params object that cleared on every call to fit (but keeping some hyperparmeters like regularisation etc).

## Further ToDo's
This PR is work in progress. I just want to line out my ideas to the community as soon as possible and enter discussion as soon as possible.  
When going through basically the whole code base I noticed some possible improvements and / or bugs related as well as unrelated to the restructuring:  
  - The `_score` function that applies the category mapping and applies random noise to training data is copied 4 times. This can be moved to a mixin
  - BaseNEncoder and SummaryEncoder are based on other encoders and do not fit into the schema proposed. Probably they can be re-wirtten to use `_fit` and `_transform` functions of the underlying encoders rather than their `fit` and `transform`. This way the main encoder would still be in charge of handling inputs, dropping invariants etc. I'll look into this.
  - Docstrings: While most docstrings were copy-paste, some do contain valuable information. Probably the easiest is to just create a function that overwrites the base-class function, add the docstring and only do `return super().function()`
  - Docstrings (II): In transform functions often the output shape is given, I'm not sure if this is correct for all encoders since there are some copy-paste error. However this information would be useful to have (e.g. if one input column is transformed to 1 or N output columns). I'd move this to a separate issue
  - UPDATE 2022-05-31: Resolve merge conflicts -> done
  - UPDATE 2022-05-31: Integrate Ben Reininger's pull request

## NB
Since this is a rather big change I'm not comfortable to do this on my own and I'd like to get some feedback on it from the community. I understand @wdm0006 does not have a lot of time but I'd be great if you could give your two cents on the idea of the change (without working through all the details - which would of course also be welcome).  
I'd also be interested in feedback from recent commiters and people interested in the project, e.g. @cmougan @bmreiniger


bmreiniger:
I had also been working on this.  I focused more on the sklearn tagging system suggested in #166, but I think in the tests it would be more natural to create the list of supervised encoders by checking inheritance (in my version, just recycling `TransformerWithTargetMixin` as a supervised encoder class).  (Tags may be useful for more instance-dependent tags, like `allow_nan` which would depend on `handle_missing`.)

I do think it might be nicer to leave unsupervised encoders without their own class: inheriting from the base class alone should be enough?

The other main difference (in a first look) in my implementation is that I kept each encoder's `fit` method and just refactored (or planned to) chunks into separate methods of the base encoder(s).

It looks like python 3.5 doesn't understand the type annotations.

PaulWestenthanner:
Thanks for sharing you code. I've read through it. I think we haven't done too much work twice, since I started with the structure and you with adding the metadata and parametrising tests.  
We should decide on a way to go forward. I very much like the idea of tagging encoders and parametrising tests using the tags.  
Regarding the structure and flexibility: In general I'm a fan of "Composition over Inheritance"  and hence using a flat rather than a deep inheritance structure with MixIns. As far as I know (but I'm not an expert) this is also what sklearn does. I agree that the structure does not seem very flexible, however I was planning to split the big MixIns into smaller ones. This would probably end up somewhere close to what you describe.  
Regarding this 
> I do think it might be nicer to leave unsupervised encoders without their own class: inheriting from the base class alone should be enough?  

I'm not quite sure: The way I implemented it (which is definitely up for discussion) is moving the `transform` method to the MixIns rather than the `BaseEncoder`. This did not feel quite right in the first place, but it works really well. So there has to be something if the `BaseEncoder` does not come with a `transform` method. Also why would you want to have an extra class for supervised but not for unsupervised? Shouldn't it be symmetric? The only difference is the signatures of some methods.   

In general I think we're heading into a similar direction and merging our approaches will be fruitful.  
And yes, the python 3.5 is really annoying :D I've check it supports type hints, but probably does not yet support the synthax for declaring abstract class variables.

bmreiniger:
> Also why would you want to have an extra class for supervised but not for unsupervised? Shouldn't it be symmetric? The only difference is the signatures of some methods.

An unsupervised encoder basically is just an encoder.  It's the supervised ones that need some additional handling.  Your unsupervised transform overlaps a lot with the supervised one, and refactoring them into the base encoder class makes sense to me; and after doing that, there's nothing really left for the unsupervised class to do.

category_encoders/utils.py
```diff
@@ -175,6 +182,162 @@ def get_generated_cols(X_original, X_transformed, to_transform):
...
+            raise ValueError('Must train encoder before it can be used to transform data.')
+
+        # then make sure that it is the right size
+        if X.shape[1] != self._dim:
+            raise ValueError('Unexpected input dimension %d, expected %d' % (X.shape[1], self._dim, ))
+
+    def _drop_invariants(self, X: pd.DataFrame, override_return_df: bool) -> Union[np.ndarray, pd.DataFrame]:
+        if self.drop_invariant:
+            for col in self.invariant_cols:
+                X.drop(col, 1, inplace=True)
```
jona-sassenhagen, line 256:
```suggestion
            X = X.drop(columns=self.invariant_cols)
```

The current option raises a FutureWarning in current Pandas versions. And inplace is usually frowned upon. Also this is one line shorter, and a lot faster.

category_encoders/cat_boost.py
```diff
@@ -288,7 +183,11 @@ def _transform(self, X_in, y, mapping=None):
                 X.loc[is_unknown_value, col] = np.nan
 
             if self.handle_missing == 'value':
-                X.loc[is_nan & unseen_values.isnull().any(), col] = self._mean
+                # only set value if there are actually missing values.
+                # In case of pd.Categorical columns setting values that are not seen in pd.Categorical gives an error.
+                nan_cond = is_nan & unseen_values.isnull().any()
+                if nan_cond.any():
+                    X.loc[is_nan & unseen_values.isnull().any(), col] = self._mean
```
jona-sassenhagen, line 190:
```suggestion
                nan_cond = is_nan & unseen_values.isnull().any()
                if nan_cond.any():
                    X.loc[nan_cond, col] = self._mean
```

category_encoders/cat_boost.py
```diff
@@ -297,19 +196,16 @@ def _transform(self, X_in, y, mapping=None):
...
 
-        """
+        codes[codes == -1] = len(categories)
+        categories = np.append(categories, np.nan)
 
-        if not isinstance(self.feature_names, list):
-            raise ValueError('Must fit data first. Affected feature names are not known before.')
-        else:
-            return self.feature_names
+        return_map = pd.Series(dict([(code, category) for code, category in enumerate(categories)]))
```
jona-sassenhagen, line 208: Isn't this just

```suggestion
        return_map = pd.Series(dict(enumerate(categories)))
```

category_encoders/count.py
```diff
@@ -129,104 +129,46 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False,
...
-                raise ValueError('Columns to be encoded can not contain null')
+    def _transform(self, X):
+        """Perform the transform count encoding."""
+        for col in self.cols:
 
-        if self._dim is None:
-            raise ValueError(
-                'Must train encoder before it can be used to transform data.'
-            )
+            X[col] = X.fillna(value=np.nan)[col]
```
jona-sassenhagen, line 140:
```suggestion
            X[col] = X[col].fillna(value=np.nan)
```

should be much more efficient no?

category_encoders/utils.py
```diff
@@ -175,6 +182,162 @@ def get_generated_cols(X_original, X_transformed, to_transform):
...
+        # else:
+        X_temp = self.transform(X, override_return_df=True)  # for finding invariant columns transform as if it was the test set
+        self.feature_names = X_temp.columns.tolist()
+
+        # drop all output columns with 0 variance.
+        if self.drop_invariant:
+            generated_cols = get_generated_cols(X, X_temp, self.cols)
+            self.invariant_cols = [x for x in generated_cols if X_temp[x].var() <= 10e-5]
+            try:
+                [self.feature_names.remove(x) for x in self.invariant_cols]
```
jona-sassenhagen, line 232:
```suggestion
                for x in self.invariant_cols:
                    self.feature_names.remove(x)
```

don't use a list comp for its side effects

jona-sassenhagen:
This looks like a great PR! 1500 lines of code removed already, should help a lot with maintinability.

category_encoders/cat_boost.py
PaulWestenthanner: thanks!

category_encoders/binary.py
```diff
@@ -68,88 +66,5 @@ class BinaryEncoder(BaseEstimator, TransformerMixin):
...
-        -------
-        feature_names: list
-            A list with all feature names transformed or added.
-            Note: potentially dropped features are not included!
-
-        """
-
-        return self.base_n_encoder.get_feature_names()
+    encoding_relation = utils.EncodingRelation.ONE_TO_M
+    __init__ = partialmethod(BaseNEncoder.__init__, base=2)
```
bmreiniger, line 70: I think it should be fine, but have you checked that cloning and get/set params work?

PaulWestenthanner: what do you mean by cloning? I just checked get/set params works. Also tests still work

category_encoders/utils.py
```diff
@@ -175,7 +184,281 @@ def get_generated_cols(X_original, X_transformed, to_transform):
     return current_cols
 
 
+class EncodingRelation(Enum):
```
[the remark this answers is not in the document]
PaulWestenthanner: I introduced it in the first place to add the output shape of the transform method dynamically in the docstring, but didn't get it to work since docstrings are not meant to be dynamic (I guess).  
I then decided to keep it since it could be useful metadata (e.g. you could add a test that the expected number of output columns is generated)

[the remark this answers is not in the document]
PaulWestenthanner: which could be actually a useful test by the way since when working on this I discovered that there might be problems with e.g. the backward difference encoder. This adds some intercept column and I guess if you encode two columns the latter intercept column overwrites the first one. But I'll have to check. If this turns out to be the case I'll create a separate issue for it

category_encoders/cat_boost.py
```diff
@@ -295,19 +165,21 @@ def _transform(self, X_in, y, mapping=None):
 
         return X
 
-    def get_feature_names(self):
-        """
-        Returns the names of all transformed / added columns.
+    def _more_tags(self):
```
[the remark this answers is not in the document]
PaulWestenthanner: This was also introduced for "dynamic docstrings" and kept since it introduces useful information (also here you could imagine a test, also testing that the encoders where transform does not depend on y actually give the same result with and without y).  
I agree that `transform_depends_on_y` is a more suitable name

category_encoders/utils.py
```diff
@@ -175,7 +184,281 @@ def get_generated_cols(X_original, X_transformed, to_transform):
...
+            return self.feature_names
+
+    @abstractmethod
+    def _fit(self, X: pd.DataFrame, y: Optional[pd.Series], **kwargs):
+        ...
+
+
+class SupervisedTransformerMixin(sklearn.base.TransformerMixin):
+
+    def _more_tags(self):
```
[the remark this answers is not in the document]
PaulWestenthanner: I think both options are equivalent and more or less depend on taste, is it? Or is there any real difference that I'm missing? I'm also happy with either, not having a strong preference for one or the other

category_encoders/binary.py
bmreiniger: `sklearn.base.clone` gets used to copy an estimator, keeping hyperparameters but dropping fitted parameters.  It uses `get_params` and then instantiates a new class of the same type using the gotten parameters, so I expect this is fine if `get_params` works.

category_encoders/utils.py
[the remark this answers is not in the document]
bmreiniger: I think you're right that they're equivalent, I just think inheritance is simpler to understand, being base python instead of an sklearn mechanism.

PR #334, opened 2022-01-18
Fix pandas future warning for dropping invariants

jona-sassenhagen:
When I use `drop_invariant=True`, I get a `FutureWarning` because `df.drop` is switching to named kwargs only. This fixes it.

It should also be substantially more performant as the DF is only modified once.

### DIFF
--- a/category_encoders/backward_difference.py
+++ b/category_encoders/backward_difference.py
@@ -214,7 +214,7 @@ class BackwardDifferenceEncoder(BaseEstimator, TransformerMixin):
         X = self.backward_difference_coding(X, mapping=self.mapping)
 
         if self.drop_invariant:
-            X = X.drop(columns=colself.drop_cols)
+            X = X.drop(columns=self.drop_cols)
 
         if self.return_df or override_return_df:
             return X
--- a/category_encoders/backward_difference.py
+++ b/category_encoders/backward_difference.py
@@ -214,8 +214,7 @@ class BackwardDifferenceEncoder(BaseEstimator, TransformerMixin):
         X = self.backward_difference_coding(X, mapping=self.mapping)
 
         if self.drop_invariant:
-            for col in self.drop_cols:
-                X.drop(col, 1, inplace=True)
+            X = X.drop(columns=colself.drop_cols)
 
         if self.return_df or override_return_df:
             return X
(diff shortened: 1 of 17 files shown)


category_encoders/backward_difference.py
```diff
@@ -214,8 +214,7 @@ def transform(self, X, override_return_df=False):
         X = self.backward_difference_coding(X, mapping=self.mapping)
 
         if self.drop_invariant:
-            for col in self.drop_cols:
-                X.drop(col, 1, inplace=True)
+            X = X.drop(columns=self.drop_cols)
```
zachmayer, line 217: Why'd you remove `inplace=True` here?

jona-sassenhagen: I think the pandas dev team is not very happy with `inplace` and there's semi-regular attempts to get rid of it in general. I personally think inplace is not the most readable and transparent way to write pandas code and I try to avoid it.

There's an extensive discussion here: https://github.com/pandas-dev/pandas/issues/16529

jona-sassenhagen:
@janmotl any chance you can review this?

janmotl:
@jona-sassenhagen Can you confirm that the new call also works in the oldest version of Pandas that category_encoders claim to support? If yes, I do not see a reason why the PR should not be merged. 

However, I am not active in the project anymore and the merge must be done by some of the current maintainers. @zachmayer ?

jona-sassenhagen:
Yes, the new call should work, there was no change to `inplace` since .21: https://pandas.pydata.org/pandas-docs/version/0.21/generated/pandas.DataFrame.drop.html

Who are the current maintainers? I don't want to randomly ping people. Zach isn't a maintainer I'm pretty sure :)

Thanks!

janmotl:
@jona-sassenhagen I don't know.

PaulWestenthanner:
Hi Jona,
thanks for your contribution. I'm one of the maintainers. I'm sorry it always takes some time until one of us reviews stuff.  
The PR looks good to me and I'll merge it when the pipeline succeeds.

PaulWestenthanner:
It seems like our github runner instance does not support Python 3.5 anymore. So the problem seems to be on our side, not on yours. 
I thinks we should deprecate python 3.5 anyways since it ran out of the 5 year support period. Please bear with me with this issue.

jona-sassenhagen:
Sure no worries. Thanks a lot.

PaulWestenthanner:
@jona-sassenhagen I've updated the pipeline. Can you please pull the latest master so that I can run the checks again?

jona-sassenhagen:
@PaulWestenthanner like this?

jona-sassenhagen:
🙏

Issue #311, opened 2021-08-02
Cat_boost encoder 


xuanqing94:
https://github.com/scikit-learn-contrib/category_encoders/blob/a810a4b7abfce9fc4eb7fc401e3d37f2c1c6e402/category_encoders/cat_boost.py#L278

That does't look like an interpolation of mean of global vs. mean of subgroup. I think it should be something like:

```python
 level_means = ((colmap['sum'] + self._mean * self.a) / (colmap['count'] + self.a)).where(level_notunique, self._mean) 
```

GLevV:
Yep, [this commit](https://github.com/scikit-learn-contrib/category_encoders/commit/25accdc9b5b1998ae317e2bcd120464a7b1027bf) forget to add multiplication by a. It's a bug, good catch.

[Reference papaer, eq 1](http://learningsys.org/nips17/assets/papers/paper_11.pdf)

PaulWestenthanner:
fixed by #339 
Thanks :)

PR #339, opened 2022-02-16
Bugfix

GLevV:
fixed bug when CatBoostEncoder did not use _a_ parameter to weight the prior #311


PaulWestenthanner:
Thanks for the fix @GLevV and for pointing it out @xuanqing94

PR #363, opened 2022-07-12
Multiclass target encoder

nercisla:
Implement fit and transform function of target encoding for multi-class target.

## Proposed Changes

  - For multi-class target y, perform one-hot encoding on y to obtain n binary columns, where n = len(y.unique()).
  - Fit each categorical column in X to each of the n one-hot encoded target columns
  - Transform the original X using the new mapping.
  - Therefore in the case of multi-class targets, transformed X categorical columns will each have len(y.unique()) columns

### DIFF
--- a/category_encoders/target_encoder.py
+++ b/category_encoders/target_encoder.py
@@ -119,6 +119,8 @@ class TargetEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
             y = ohe_encoder.transform(y.astype(str))
             self.y_colnames = y.columns
             self.X_colnames = [col + "_" + ycol for ycol in y.columns for col in self.cols]
+        else:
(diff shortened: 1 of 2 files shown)


PaulWestenthanner:
Is this the same thing that the `PolynomialWrapper` (not the obvious choice of name, sorry) does in a generic way for all encoders, so it is not needed specifically for target encoding? It pretty much looks like it, but I might miss something

nercisla:
I think you are right, Paul, thanks!  I never used this wrapper - possibly due to its name.  I think I prefer the return of all encoded columns, but I can't see a justifiable use case for it. In light of this, a slightly unusable piece of code...

PaulWestenthanner:
Alright. thanks for your work anyway and sorry again for the confusing naming

PR #366, opened 2022-08-07
Target encoding hierarchical

### LINKED ISSUE #136: Implement Target Encoding with Hierarchical Structure Smoothing
From section 4 of the [paper](https://kaggle2.blob.core.windows.net/forum-message-attachments/225952/7441/high%20cardinality%20categoricals.pdf) sited in TargetEncoding.

> Instead of choosing the prior probability of the target as the _null hypothesis_, it is reasonable to replace it with the estimated probability at the next higher level of aggregation in the attribute hierarchy

In other words, if we have a single zipcode 54321 but 100 zipcodes 54322 and 100 zipcodes 54323, we could use the mean of zipcode level 4 _5432X_ as our mean smoothing term for zipcode 54321, instead of the mean for all zipcodes XXXXX.

This would be a really nice additional piece of functionality to add as an encoder.

PaulWestenthanner:
Fixes #136 

## Proposed Changes
This pull request implements feature hierarchies in Target Encoders.  
Author: @nercisla
Current status: Work in Progress


PaulWestenthanner:
Hi @nercisla 

Thanks for your work. I created a draft pull request just because it is easier and more transparent to make comments and have the tests run. I hope that's fine with you. Also feel free to change the description if you want to

tests/test_target_encoder.py
```diff
@@ -112,3 +112,67 @@ def test_HandleUnknownValue_HaveUnknownInTest_ExpectMean(self):
         obtained = enc.transform(test, test_target)
 
         self.assertEqual(.6, list(obtained['color'])[0])
+
```
PaulWestenthanner, line 115: another nice test would be to check with a "trivial hierarchy", i.e. mapping everything to a unique top level does the same thing as doing target encoding without any hierarchy

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
...
+                if self.heirarchy and col in self.heirarchy:
+                    col_heir = 'HEIR_'+col
+                    stats_heir = y.groupby(X[col_heir]).agg(['count', 'mean'])
+                    smoove_heir = self._weighting(stats_heir['count'])
+                    scalar_heir = scalar * (1 - smoove_heir) + stats_heir['mean'] * smoove_heir
+                    scalar_heir_long = X[[col, col_heir]].drop_duplicates()
+                    scalar_heir_long.index = np.arange(1, scalar_heir_long.shape[0]+1)
+                    scalar = scalar_heir_long[col_heir].map(scalar_heir.to_dict())
+                    X.drop([col_heir], axis=1, inplace=True)
+                    # Added line to reduce dimension of ordinal encoder - Whats the third dimension?
```
PaulWestenthanner, line 147: what do you mean by "third dimension"? The dimension in the ordinal encoder are the number of columns that are encoded. So subtracting 1 here should be fine

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
...
+            if 'HEIR_' not in col:
+                values = switch.get('mapping')
 
-            prior = self._mean = y.mean()
+                scalar = prior
+                if self.heirarchy and col in self.heirarchy:
+                    col_heir = 'HEIR_'+col
+                    stats_heir = y.groupby(X[col_heir]).agg(['count', 'mean'])
+                    smoove_heir = self._weighting(stats_heir['count'])
+                    scalar_heir = scalar * (1 - smoove_heir) + stats_heir['mean'] * smoove_heir
```
PaulWestenthanner, line 142: I would find this more readable if you write `prior` instead of `scalar`. This should be the same right?

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
...
         for switch in self.ordinal_encoder.category_mapping:
             col = switch.get('col')
-            values = switch.get('mapping')
+            if 'HEIR_' not in col:
+                values = switch.get('mapping')
 
-            prior = self._mean = y.mean()
+                scalar = prior
+                if self.heirarchy and col in self.heirarchy:
+                    col_heir = 'HEIR_'+col
```
PaulWestenthanner, line 139: this also seems to work for single-level hierarchies only, right? So at some point we'd need to raise an exception (probably when parsing the hierarchy map)

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
...
+                    stats_heir = y.groupby(X[col_heir]).agg(['count', 'mean'])
+                    smoove_heir = self._weighting(stats_heir['count'])
+                    scalar_heir = scalar * (1 - smoove_heir) + stats_heir['mean'] * smoove_heir
+                    scalar_heir_long = X[[col, col_heir]].drop_duplicates()
+                    scalar_heir_long.index = np.arange(1, scalar_heir_long.shape[0]+1)
+                    scalar = scalar_heir_long[col_heir].map(scalar_heir.to_dict())
+                    X.drop([col_heir], axis=1, inplace=True)
```
PaulWestenthanner, line 146: we had a discussion in this project recently where a pandas guy discouraged using `inplace` and actually removed all (or at least most of) `inplace` usages from this repo. I don't remember the exact reason but it made sense to me and we should probably follow suit here as well. So just use  
`X = X.drop(...)`

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
...
 
-            mapping[col] = smoothing
+                if self.handle_missing == 'return_nan':
+                    smoothing.loc[values.loc[np.nan]] = np.nan
+                elif self.handle_missing == 'value':
+                    smoothing.loc[-2] = prior
+
+                mapping[col] = smoothing
+
+        self.ordinal_encoder.mapping = [map for map in self.ordinal_encoder.mapping if not "HEIR_" in map['col']]
```
PaulWestenthanner, line 170: This is basically cleaning the ordinal encoder so that it never new there was hierarchy involved? I like the idea that all the additional columns should be hidden from the user

tests/test_target_encoder.py
PaulWestenthanner, line 115: and maybe another test for multi-level hierarchies where we'd expect an error.

PaulWestenthanner (commented):
Thanks a lot for your work!  
I've added some suggestions. As a general remark some tests are still failing, we'd need to fix them. Also I'm curious what the `heirarchical_example.numbers` file is about and whether it needs to be added to the repository

category_encoders/target_encoder.py
```diff
@@ -83,7 +83,7 @@ class TargetEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
     encoding_relation = util.EncodingRelation.ONE_TO_ONE
 
     def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, handle_missing='value',
-                 handle_unknown='value', min_samples_leaf=1, smoothing=1.0):
+                 handle_unknown='value', min_samples_leaf=1, smoothing=1.0, heirarchy=None):
```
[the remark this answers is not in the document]
nercisla: Yes, you are correct.  Though the word heir is spelt with i and e the opposite way around. ;)

category_encoders/target_encoder.py
```diff
@@ -99,8 +99,22 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
                           category=FutureWarning)
         self.mapping = None
         self._mean = None
+        self.heirarchy = heirarchy
+        self.cols_heir = []
 
     def _fit(self, X, y, **kwargs):
+        # Is this needed?
+        X = X.copy()
```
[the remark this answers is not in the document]
nercisla: No, it's not needed.  It was a remnant of an old solution that still required checking.

category_encoders/target_encoder.py
```diff
@@ -99,8 +99,22 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
+
+        if self.heirarchy:
+            for switch in self.heirarchy:
+                if switch in X.columns:
+                    new_column = 'HEIR_'+switch
+                    X[new_column] = X[switch].map(self.heirarchy[switch])
+                    self.cols_heir.append(new_column)
+
+        # self.cols not declared before here? X Has all required columns ------
```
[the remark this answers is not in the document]
nercisla: This is also be a remnant of an old solution.

category_encoders/target_encoder.py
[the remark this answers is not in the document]
nercisla: I have added this now.

category_encoders/target_encoder.py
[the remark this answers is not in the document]
nercisla: Working on fixing the tests, Joe and I found a problem here, so the method is a little different now.  We separate the OrdinalEncoders for the base column(s) and the hierarchy.

tests/test_target_encoder.py
```diff
@@ -112,3 +112,67 @@ def test_HandleUnknownValue_HaveUnknownInTest_ExpectMean(self):
...
+            {
+                'Compass': ['N', 'N', 'NE', 'NE', 'NE', 'SE', 'SE', 'S', 'S', 'S', 'S', 'W', 'W', 'W',
+                            'W', 'W'],
+                'Speed': ['s', 's', 's', 's', 'm', 'm', 'm', 'f', 'f', 'f', 'f', 'f', 'f', 'f', 'f', 'f'],
+                'Animal': ['Cat', 'Cat', 'Cat', 'Cat', 'Cat', 'Dog', 'Dog', 'Dog', 'Dog',
+                           'Dog', 'Dog', 'Tiger', 'Tiger', 'Wolf', 'Wolf', 'Cougar'],
+                'target': [1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 0, 1, 0, 0, 0, 1]
+            }, columns=['Compass', 'Speed', 'Animal', 'target'])
+
+        heirarchical_cat_example = pd.DataFrame(
```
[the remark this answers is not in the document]
nercisla: Correct.

category_encoders/target_encoder.py
```diff
@@ -99,8 +99,22 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
 
     def _fit(self, X, y, **kwargs):
+        # Is this needed?
+        X = X.copy()
+
+        if self.heirarchy:
+            for switch in self.heirarchy:
+                if switch in X.columns:
+                    new_column = 'HEIR_'+switch
+                    X[new_column] = X[switch].map(self.heirarchy[switch])
```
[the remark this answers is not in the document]
nercisla: At this stage, we only consider one level of hierarchy, but know how to change it, so can do that soon.  Either this release or another.

category_encoders/target_encoder.py
```diff
@@ -113,34 +127,52 @@ def _fit(self, X, y, **kwargs):
 
     def fit_target_encoding(self, X, y):
         mapping = {}
+        prior = self._mean = y.mean()
 
         for switch in self.ordinal_encoder.category_mapping:
             col = switch.get('col')
-            values = switch.get('mapping')
+            if 'HEIR_' not in col:
```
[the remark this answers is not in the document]
nercisla: Agreed.

tests/test_target_encoder.py
nercisla: Yup, agreed, as this currently breaks, I think.

category_encoders/target_encoder.py
nercisla: No, I don't believe that is true.  The prior is taken over all the data, but the scalar is an estimate of the probability.  The paper clearly defines it separately, so feel the language should follow.

category_encoders/target_encoder.py
nercisla: Agreed.

category_encoders/target_encoder.py
nercisla: No problem.  Though nicely this line has been dropped due to the cleaned up methodology :)

category_encoders/target_encoder.py
nercisla: As above, this line has been removed in the new method.

tests/test_target_encoder.py
nercisla: Do you mean a test to check it errors elegantly?

PaulWestenthanner: yes, just an `self.assertRaises` statement that checks for the error of mulit-level hierarchies

category_encoders/target_encoder.py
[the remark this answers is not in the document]
PaulWestenthanner: I'm fine with having it in another release. Let's get the first working version merged soon and then iterate from there

tests/test_target_encoder.py
PaulWestenthanner: I just saw you've already implemented that. Very nice!

category_encoders/target_encoder.py
PaulWestenthanner: The variable called `scalar_heir` is defined separately and should keep its name. My comment war referring to the variable called `scalar` on the same line. This is even set equal to `prior` in line 155.

tests/test_target_encoder.py
PaulWestenthanner: Why do you think this breaks? According to the paper this should be fine, right?

PaulWestenthanner:
Hi @nercisla 
thanks for updating the PR. I think it looks really good now. The only open issue I see is the test that checks if a trivial hierarchy yields the same encoding as no hierarchy.  
Do you see any other open topics that need to be fixed before merging?
Really looking good now and I think we can merge next week. Thank you!

category_encoders/target_encoder.py
nercisla: So, the scalar is normally set to the prior, but for multiple hierarchies, the scalar is a function.  This matters more when we will have multi-level hierarchies, in which the "scalar" will definitely move away from the prior.  So, in that sense, I think it matters that we don't call it the prior.  The prior and scalar will be different and that also matters for the default values (-1 and -2) in the mapping.

nercisla:
I think the trivial hierarchy issue is tested by two tests:

- test_hierarchy_single_mapping - tests a hierarchy that (with the given data) results in a single mapped value.  It should return the original without a mapping
- test_hierarchy_no_mapping - tests a hierarchy that maps the same as the level below (for the given data).  It should again return the original without a mapping

nercisla:
The features we want to move on to are (in order of priority in our opinion):

1. User gives a column that contains the mapped higher level (i.e. they have already applied a custom function to a column to get the next level up)
2. Multi-level hierarchies
3. Typical hierarchical functions (such as Zipcode/postcode etc)

We think these are "nice to have" extras, which don't need to be included just now.

nercisla:
For info, I have a concern over the mapping.  The current code works and gives the right answer, but the mapping is reversed if we want to consider multi-level mapping.  So it might be good to consider larger hierarchies and a solution before we merge...

Detail: we currently have a mapping dictionary that goes from the existing column (key) to the higher level mapping (value).  But for multi-level, it makes more sense to have the highest levels first in the dictionary.  So, we need to consider reversing the map.  The mapping won't be so trivial for multi-level in that scenario, but makes more sense for the user.

nercisla:
Joe and I have modified the code to "reverse" the mapping dictionary.  This should now follow a more user-friendly version of mapping.   Multi-level mapping is also complete and we've included a test.

category_encoders/target_encoder.py
```diff
@@ -104,22 +105,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
-        self.hierarchy = hierarchy
+            self.hierarchy = {}
+            self.hierarchy_depth = {}
+            for switch in hierarchy:
+                D = fd.flatten(hierarchy[switch], inverse=True)
+                hierarchy_check = self._check_dict_key_tuples(D)
+                self.hierarchy_depth[switch] = hierarchy_check[1]
+                if not hierarchy_check[0]:
+                    raise ValueError('Hierarchy mapping contains different levels for key "' + switch + '"')
+                self.hierarchy[switch] = {(k if type(t) is tuple else t): v for t, v in D.items() for k in t}
```
PaulWestenthanner, line 116: checking for types should be done using `isinstance(t, tuple)`

category_encoders/target_encoder.py
```diff
@@ -104,22 +105,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
 
     def _fit(self, X, y, **kwargs):
         if self.hierarchy:
             X_hier = pd.DataFrame()
             for switch in self.hierarchy:
                 if switch in self.cols:
-                    new_column = 'HIER_'+str(switch)
-                    X_hier[new_column] = X[str(switch)].map(self.hierarchy[str(switch)])
-                    self.cols_hier.append(new_column)
+                    colnames = ['HIER_' + str(switch) + '_' + str(i+1) for i in range(self.hierarchy_depth[switch])]
```
PaulWestenthanner, line 131: an f-string would be more readable here

PaulWestenthanner:
Great! Thanks for implementing the multilevel hierarchy. Maybe it would be nice to give an example of the hierarchy parameter in the docstrings since this is getting a little complicated (especially with multilevel)

category_encoders/target_encoder.py
```diff
@@ -4,6 +4,7 @@
 from category_encoders.ordinal import OrdinalEncoder
 import category_encoders.utils as util
 import warnings
+import flatten_dict as fd
```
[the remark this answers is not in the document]
PaulWestenthanner: this is also the reason why the tests are failing at the moment

tests/test_target_encoder.py
PaulWestenthanner: I just tried it out and it works fine. Here's a complete test. I'd be happy if you include it

```python
    def test_trivial_hierarchy(self):
        trivial_hierarchical_map = {
            'Plant': {
                'Plant': ('Rose', 'Daisy', 'Daffodil', 'Bluebell')
            }
        }

        enc_hier = encoders.TargetEncoder(verbose=1, smoothing=2, min_samples_leaf=2, hierarchy=trivial_hierarchical_map,
                                     cols=['Plant'])
        result_hier = enc_hier.fit_transform(self.hierarchical_cat_example, self.hierarchical_cat_example['target'])
        enc_no_hier = encoders.TargetEncoder(verbose=1, smoothing=2, min_samples_leaf=2, cols=['Plant'])
        result_no_hier = enc_no_hier.fit_transform(self.hierarchical_cat_example, self.hierarchical_cat_example['target'])
        pd.testing.assert_series_equal(result_hier["Plant"], result_no_hier["Plant"])
```

category_encoders/target_encoder.py
PaulWestenthanner: you're right! thanks for that explanation

category_encoders/target_encoder.py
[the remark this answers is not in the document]
nercisla: I think the flatten-dict function is complete and possibly too extensive to re-write.  We could ask the author if we could copy the required function into our utils, but I'd hesitate to re-write it.  We tried and found flatten-dict to be better.  Will add it into setiup.py for the moment, until we reach concensus.

category_encoders/target_encoder.py
```diff
@@ -104,22 +105,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
         self.mapping = None
         self._mean = None
         if hierarchy:
-            if self._depth(hierarchy) > 2:
-                raise ValueError('Hierarchy mapping contains too many levels')
-        self.hierarchy = hierarchy
+            self.hierarchy = {}
+            self.hierarchy_depth = {}
+            for switch in hierarchy:
+                D = fd.flatten(hierarchy[switch], inverse=True)
```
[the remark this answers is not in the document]
nercisla: Will change this!

category_encoders/target_encoder.py
```diff
@@ -104,22 +105,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
+        else:
+            self.hierarchy = hierarchy
         self.cols_hier = []
 
-    def _depth(self, d):
-        return max(self._depth(v) if isinstance(v, dict) else 0 for v in d.values()) + 1
+    def _check_dict_key_tuples(self, d):
+        min_tuple_size = min(len(v) for v in d.values())
+        max_tuple_size = max(len(v) for v in d.values())
+        return True if min_tuple_size == max_tuple_size else False, min_tuple_size
```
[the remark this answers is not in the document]
nercisla: > also in the `True` case you want to return the size right? otherwise `hierarchy_check[1]` might be undefined (c.f. line 113)

This actually does the job despite what it seems at first glance.  The if/else come first then the min_size is added to either afterwards.  So it still returns the min_size in both cases.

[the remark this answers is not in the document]
nercisla: But I agree it's not easily readable (even for me second time round) so will change it.

category_encoders/target_encoder.py
nercisla: Agreed.

nercisla:
Have added a simple example, since more complex ones might take even more space.

category_encoders/target_encoder.py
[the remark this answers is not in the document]
PaulWestenthanner: aaah that's how it works. Thanks for changing anyway

category_encoders/target_encoder.py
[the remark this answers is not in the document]
PaulWestenthanner: I think I just found a neat way:  
pandas apparently implements a function `json_normalize` that can be used to flatten dicts as well. This is very well maintained. For reference it's way number 3 here: https://www.freecodecamp.org/news/how-to-flatten-a-dictionary-in-python-in-4-different-ways/
I just tried it out and it seems to work:  

With this function
```python
def flatten_reverse_dict(d):
    sep = "___"
    [flat_dict] = pd.json_normalize(d, sep=sep).to_dict(orient='records')
    reversed_flat_dict = {v: tuple(k.split(sep)) for k,v in flat_dict.items()}
    return reversed_flat_dict
```  
we get the same results as otherwise  
```python
hierarchy_multi_level = {
    'Animal': {
        'Warm-Blooded':
            {'Mammals': ('Cat', 'Dog'),
             'Birds': ('Osprey', 'Kite'),
             'Fish': ('Carp', 'Clownfish')
             },
        'Cold-Blooded':
            {'Reptiles': ('Lizard'),
             'Amphibians': ('Snake', 'Frog')
             }
    }}

flattend = flatten(hierarchy_multi_level["Animal"], inverse=True)
flattend_pd = flatten_reverse_dict(hierarchy_multi_level["Animal"])
print(flattend_pd == flattend)  # prints True
``` 
This would be just 4 lines.  
What do you think? Is this an acceptable solution for you?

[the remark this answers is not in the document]
nercisla: A very nice solution!  I have tested it, so will push it to the branch now.  Nice work!

PaulWestenthanner:
Great!  
Thank you very much @nercisla and @joeybricks  
The tests are succeeding and I'd consider it ready for the merge. Is this fine with you or is there anything missing?

nercisla:
Very happy to merge!

PR #372, opened 2022-09-29
Include dict comprehension

boneyag:
## Proposed Changes

Replace `dict` constructor with `dict` comprehension In `category_encoders/leave_one_out.py`, because it is much faster.


PaulWestenthanner:
Thanks for pointing that out and fixing it :)

PR #373, opened 2022-10-04
Target encoding heirarchical columnwise

nercisla:
This pull request enhances hierarchies in Target Encoders.

Author: @nercisla
Current status: Work in Progress

## Proposed Changes

Allows a user to submit a hierarchy within a dataframe (i.e. columnwise), not just a mapping dictionary.
Columns must take the names HIER_colA_1, HIER_colA_2, HIER_colA_3, HIER_colB_1, HIER_colB_2. HIER_colC_1 etc where the last digit represents the level of hierarchy (top =1)


nercisla:
Just seen one test failed... Am sure this one worked on my Mac, so will go back and check again!

category_encoders/target_encoder.py
```diff
@@ -76,15 +76,22 @@ class TargetEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
...
-    >>> X = ['N', 'N', 'NE', 'NE', 'NE', 'SE', 'SE', 'S', 'S', 'S', 'S', 'W', 'W', 'W', 'W', 'W']
-    >>> hierarchical_map = {'Compass': {'N': ('N', 'NE'), 'S': ('S', 'SE'), 'W': 'W'}}
-    >>> y = [1, 0, 1, 1, 1, 0, 0, 1, 0, 1, 0, 1, 0, 0, 0, 1]
-    >>> enc = TargetEncoder(hierarchy=hierarchical_map).fit(X, y)
-    >>> hierarchy_dataset = enc.transform(X)
-    >>> print(hierarchy_dataset[0].values)
-    [0.5        0.5        0.94039854 0.94039854 0.94039854 0.13447071
-    0.13447071 0.5        0.5        0.5        0.5        0.40179862
-    0.40179862 0.40179862 0.40179862 0.40179862]
+    >>> X, y = load_compass()
```
PaulWestenthanner, line 79: maybe add the corresponding import statement to make it a self contained example
``` >>> from category_encoders.datasets import load_compass``

File: category_encoders/target_encoder.py [hunk 2]
```diff
@@ -113,19 +120,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
                           category=FutureWarning)
         self.mapping = None
         self._mean = None
-        if hierarchy:
+        if isinstance(hierarchy, (dict, pd.DataFrame)) and cols is None:
+            raise ValueError('Hierarchy is defined but no columns are named for encoding')
```
PaulWestenthanner, line 124: why do we need this error? the current behaviour is that if you do not specify `cols` all categorical columns will be determined and encoded. So if i rely on this default and not specify columns this will break now right?

PaulWestenthanner: ah ok, in line 138 you have to know at init time what the columns will look like, determining them on the fly during fit-time will not be enough

category_encoders/target_encoder.py
```diff
@@ -113,19 +120,33 @@ def __init__(self, verbose=0, cols=None, drop_invariant=False, return_df=True, h
...
                     raise ValueError('Hierarchy mapping contains different levels for key "' + switch + '"')
                 self.hierarchy[switch] = {(k if isinstance(t, tuple) else t): v for t, v in flattened_hierarchy.items() for k in t}
-        else:
+        elif isinstance(hierarchy, pd.DataFrame):
+            self.hierarchy = hierarchy
+            self.hierarchy_depth = {}
+            for col in self.cols:
+                HIER_cols = self.hierarchy.columns[self.hierarchy.columns.str.startswith(f'HIER_{col}')].values
+                HIER_levels = [int(i.replace(f'HIER_{col}_', '')) for i in HIER_cols]
+                if np.array_equal(sorted(HIER_levels), np.arange(1, max(HIER_levels)+1)):
```
PaulWestenthanner, line 141: I think we should document somewhere that in case of supplying a dataframe as hierarchy that the columns have to be called HIER_col_1, ... HIER_col_N.  
Also I'm not sure if 1 or N is the coarsest level (so probably also worth documenting)

category_encoders/target_encoder.py
```diff
@@ -134,14 +155,19 @@ def _check_dict_key_tuples(self, d):
         return min_tuple_size == max_tuple_size, min_tuple_size
 
     def _fit(self, X, y, **kwargs):
-        if self.hierarchy:
+        if isinstance(self.hierarchy, dict):
+            if not set(self.cols).issubset(X.columns):
```
PaulWestenthanner, line 159: this check is not specific to hierarchy right? if so it should be moved  to the base encoder. Strange that we do not have such a check implemented already...

category_encoders/target_encoder.py
PaulWestenthanner: in the postcode example cols should be just `postcode` right? so specifying `cols=None` would not make any sense anyway since the HIER_postcode_N cols would be encoded as columns on their own instead of being recognized in the hierarchy.  
Then this is just fine to throw an error here

PaulWestenthanner (commented):
Very nice pull request. Thanks for your work!  
I think the code is merge-able as it is, just a little adjustment to documentation has to be made here and there :)

category_encoders/datasets/__init__.py
```diff
@@ -0,0 +1,7 @@
+from ._base import load_compass
```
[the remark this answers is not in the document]
nercisla: Yes, this is the model we copied to keep things standard :)

category_encoders/datasets/_base.py
```diff
@@ -0,0 +1,65 @@
+"""
+Base IO code for datasets
+"""
+
+import pkg_resources
+import pandas as pd
+
+def load_compass():
+    """Return a dataframe for target encoding with 100 UK postcodes and hierarchy.
```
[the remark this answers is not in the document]
nercisla: Fixed

category_encoders/datasets/_base.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Parameters
+    ----------
+    target_type : str, default='binary'
+        Options are 'binary', 'non_binary', 'categorical'
+
+    Returns
+    -------
+    X: A dataframe containing features
+    y: A dataframe containing the target variable
```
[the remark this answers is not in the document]
nercisla: Fixed

category_encoders/datasets/_base.py
```diff
@@ -0,0 +1,65 @@
...
+    Contains the following fields:
+        index                16 non-null int64
+        compass              16 non-null object
+        HIER_compass_1       16 non-null object
+        target               16 non-null int64
+
+    Returns
+    -------
+    X: A dataframe containing features
+    y: A dataframe containing the target variable
```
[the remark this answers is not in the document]
nercisla: Fixed

category_encoders/target_encoder.py
nercisla: Agreed.  We have moved this, hopefully to the right place.

PaulWestenthanner:
LGTM :)

PaulWestenthanner:
I'll also build a release now so the changes will be ready to be pip installed

Issue #351, opened 2022-04-28
Catboost fit_transform method is broken.


PraveshKoirala:
## TLDR
When called fit_transform, the output is shuffled and not in sync with the input. This does not occur when called transform() after fitting. This tidbit should either be explained in documentation or should be solved to give the expected (i.e. ordered) results (order being values in the same order as the categories).

## Example
```
cats = pd.DataFrame({'X':['a', 'a', 'd', 'd', 'd', 'f', 'f'], 'id': ['a', 'a', 'd', 'd', 'd', 'f', 'f'], 'y' : [0, 0, 1, 0, 1, 1, 1]})
encoder = ce.CatBoostEncoder(cols=['X'])
encoder.fit(cats, y, return_df=True)
encoder.transform(cats)
```
gives

![image](https://user-images.githubusercontent.com/2066100/165650004-d59c5726-1bdc-42b4-b3ae-d4b40721e0c8.png)

```
cats = pd.DataFrame({'X':['a', 'a', 'd', 'd', 'd', 'f', 'f'], 
                     'id': ['a', 'a', 'd', 'd', 'd', 'f', 'f'], 
                     'y' : [0, 0, 1, 0, 1, 1, 1]})
encoder = ce.CatBoostEncoder(cols=['X'])
encoder.fit_transform(cats, y, return_df=True)
```

gives (NOTICE DIFFERENT VALUES FOR TWO a's):
![image](https://user-images.githubusercontent.com/2066100/165649879-b8ef64a0-ecc1-4258-addb-50114a733d53.png)

PaulWestenthanner:
Hi @PraveshKoirala 

this is not a bug.
`fit_transform` calls `transform(X, y)` with the target information. As stated in catboost transform documentation

 > y : array-like, shape = [n_samples] when transform by leave one out
 >           None, when transform without target information (such as transform test set)

This always leaves out the current value. Hence we expect to see some differences. Indeed 
``` python
cats = pd.DataFrame({'X':['a', 'a', 'd', 'd', 'd', 'f', 'f'], 'id': ['a', 'a', 'd', 'd', 'd', 'f', 'f'], 'y' : [0, 0, 1, 0, 1, 1, 1]})
encoder = ce.CatBoostEncoder(cols=['X'])
encoder.fit(cats, y, return_df=True)
encoder.transform(cats, y)
```
gives the same result as `fit_transform`.
Does this make sense for you?

PaulWestenthanner:
[...]

Issue #337, opened 2022-02-11
Support has_time parameter in CatBoostEncoder


kylegilde:
## Expected Behavior
In CatBoost, the highly-cardinal categorical features can be encoded in two ways using the has_time parameter:

1.  w/ random permutation (has_time=False)
2. w/o random permutation (has_time=True)

https://catboost.ai/en/docs/concepts/parameter-tuning#internal-dataset-order
https://github.com/catboost/catboost/issues/1076#issuecomment-905039708

## Actual Behavior
The CatBoostEncoder only supports the 1st method.

## Proposal

Add the has_time parameter to CatBoostEncoder in order to support ordered datasets

GLevV:
I think it's the opposite: existing implementation of CBE uses no permutation, so it is time-aware implementation. Old implementation used LOO once, so it was 1 round of permutations.

Maybe original CBE implementation coders could correct me if I'm wrong.

kylegilde:
@GLevV , I wasn't notified when you made your comment. 

Interestingly, the [docstring indicates that the random permutation](https://github.com/scikit-learn-contrib/category_encoders/blob/master/category_encoders/cat_boost.py#L23-L26) is occurring, but I don't see that it is happening anywhere in the code itself.

I wonder if the docstring is incorrect?

PaulWestenthanner:
I'm not too familiar with out implementation of catboost encoder. I'd need to have a look at it. I cannot guarantee the docstrings are correct

PR #381, opened 2022-10-31
[DOC] Catboost docs reformulation

glevv:
Connected to #337 and #351

### DIFF
--- a/category_encoders/cat_boost.py
+++ b/category_encoders/cat_boost.py
@@ -22,8 +22,10 @@ class CatBoostEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
     does not have time dependency it should still work just fine assuming
     sorting of the data won't leak any information.
 
-    Regularization (parameter a) is achieved by adding it to running counts
-    (so called pseudocounts).
+    Regularization (parameter a) is achieved by adding this imcrement to running counts
+    (in this case so called pseudocounts).
+    Regularization (parameter sigma) is achieved by adding gaussian noise to
+    the output.
 
     NOTE: behavior of the transformer would differ in transform and fit_transform
     methods depending if y values are passed. If no target is passed then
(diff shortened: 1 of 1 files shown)


category_encoders/cat_boost.py
```diff
@@ -13,19 +13,20 @@ class CatBoostEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
...
-        perm = np.random.permutation(len(X))
-        X = X.iloc[perm].reset_index(drop=True)
-        y = y.iloc[perm].reset_index(drop=True)
-
-    This is necessary because some data sets are sorted based on the target
-    value and this coder encodes the features on-the-fly in a single pass.
+    CatBoost encoder is the variation of target encoding. It supports
+    time-aware encoding, regularization and online learning.
+
+    This implementation is time-aware (similar to CatBoos 'has_time=True'),
```
PaulWestenthanner, line 19: there is a typo heere `CatBoosts`

category_encoders/cat_boost.py
```diff
@@ -13,19 +13,20 @@ class CatBoostEncoder(util.BaseEncoder, util.SupervisedTransformerMixin):
...
+    so no random permutations are used. It makes this encoder sensitive to
+    ordering of the data and suitable for time series problems. If your data
+    does not have time dependency it should still work just fine assuming
+    sorting of the data won't leak any information outside training scope
+    (i.e. no data leakage).
+
+    NOTE: behavior of the transformer would differ in transform and fit_transform
+    methods depending if y values are passed. If no target is passed then
+    encoder will map the last value of running mean to each category. If y is passed
+    then it will continue to update running mean and encode it to passed feature columns.
```
PaulWestenthanner, line 29: I don't like the word `continue` here. The encoder is newly fitted according to target information.  
If I do something like
``` python 
enc = CatBoostEncoder()
enc.fit_transform(X1, y1)
enc.fit_transform(X2, y2)
``` 
the previous fit will be ignored. So `continue` might be misleading and `refit` might be a better wording. Do you agree?

glevv: I reworded it, I think it should be better

PaulWestenthanner:
Thanks a lot for this contribution @glevv

Issue #384, opened 2022-11-16
skl v1.2


bmreiniger:
scikit-learn will be releasing version 1.2 soon.  I upgraded to their master branch and ran the tests here; everything appears to work except that

1. sklearn will start throwing an error if feature names are not all strings ([source](https://github.com/scikit-learn/scikit-learn/blame/e70fb3f91b00f59899c63dc32cd8fc9162284b60/sklearn/utils/validation.py#L1884)).  The test `test_encoders.test_column_transformer` uses `tests.helpers.create_dataset` which includes one column with name `321`, and breaks.  I don't think there's much reason to support the mixed column name types, and suggest modifying the test dataset.
2. the `load_boston` dataset has been deprecated and will be removed in 1.2; the test `test_encoders.test_string_index` should be changed to some other sklearn dataset, or perhaps the internal one.

I'm surprised I don't see errors from `get_features_names` vs. `get_feature_names_out`, since #382 isn't merged yet; have I done something wrong in setting up my environment I wonder?

PaulWestenthanner:
Hi Ben,

I just merged #382 not sure why it wasn't a problem. Do you know when sklearn 1.2 will be released as stable? For 1 I agree that we should just delete the test if it is unsupported.  
2. should be more work since all our doctests rely on the boston housing. The deprecation message suggests california housing dataset instead. Maybe we should just switch to that, but the column names will be different as well.  

Thanks for testing against 1.2 and pointing this out! great help!

bmreiniger:
1.2 release candidate just hit PyPI, with a regular release to follow in a "week or two".

PaulWestenthanner:
I probably won't have time to replace the boston housing dataset in the next couple of weeks. Should we just put some sklearn < 1.2 in the requirements for the moment?

PR #388, opened 2022-12-24
Improve hash-encoder performance - avoid unecessary Series construction

bkhant1:
I used the hash-encoder the other day and was surprised by the time it takes for a theorically quite lightweight algorithm. 

I looked into it in a bit more detail, generating a random dataframe with 100,000 rows to measure performance. It turns out that in that case >50% of the time is spent in the `pandas.Series` constructor. [Here's my notebook](https://gist.github.com/bkhant1/b6780714831219a0b3d517849322b039) if you are curious!

## Proposed Changes

  - Returns the already-constructed list in `hash_fn` instead of `pandas.Series`, and use `result_type="expand"` in `apply` the expand the lists into a dataframe.

## Result

Here's the performance table comparision

| Dataframe #columns, #rows / #nb_components, #nb_process | `pandas.Series` | Raw `list` |
|----------|-------------|------|
| 3, 30 / 10, 4 | 5.05s | 5.23s |
| 3, 30 / 10, 1 | 2.1s | 2.08s |
| 3, 30 / 100, 1 | 2.09s | 2.21s |
| 10, 10k / 10, 4 | 7.94s | 7.59s |
| 10, 10k / 10, 1 | 3.5s | 2.31s |
| 50, 1000k / 10, 4 | 1min30s | 51.4s |
| 50, 1000k / 10, 1 | 4min8s | 2min |

For larger dataframes, it slashes the time it takes to `transform` in two. [Here's the notebook](https://gist.github.com/bkhant1/6b56b82fac513180af873bf097c4df67) those results are from.

## Follow-up

I thought I would start with this change as it's a 2-liner!

But there are still some things I would want to look into:
- the multiprocessing seems to slow down things for smaller dataframes
- there might be a way around using `apply` which is know to be slow
- there might be a way to make the communication between processes faster when multiprocessing

### DIFF
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -306,14 +306,14 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
                     else:
                         hasher.update(bytes(str(val), 'utf-8'))
                     tmp[int(hasher.hexdigest(), 16) % N] += 1
-            return pd.Series(tmp, index=new_cols)
+            return tmp
 
         new_cols = [f'col_{d}' for d in range(N)]
 
         X_cat = X.loc[:, cols]
         X_num = X.loc[:, [x for x in X.columns.values if x not in cols]]
 
-        X_cat = X_cat.apply(hash_fn, axis=1)
+        X_cat = X_cat.apply(hash_fn, axis=1, result_type='expand')
         X_cat.columns = new_cols
 
         X = pd.concat([X_cat, X_num], axis=1)


PaulWestenthanner:
Hi,
thanks for your effort and this very detailed pull request. The errors in the pipeline are my fault (connected to #384). I'll try to sort it out this week and have your PR merged as well. Looks like a great improvement already

bkhant1:
Sounds good, thanks for taking the time to review!

PaulWestenthanner:
@bkhant1  I've just merged the fixes in master. Can you please pull master and see if the pipeline still fails?

bkhant1:
Just done it - the workflow needs approval to run.

PaulWestenthanner:
LGTM! I'll merge it. Sorry again for the broken pipeline in the first place. Looking forward to see more contributions

Issue #395, opened 2023-01-21
sklearn compatibility with feature_names_out


PaulWestenthanner:
## Expected Behavior
The behaviour of `get_feature_names_out` should be the same as for sklearn

## Actual Behavior an explanation in doc.

At the moment `feature_names_out_` attribute is set in `fit` and returned by `get_feature_names_out` function. In sklearn the `get_feature_names_out` function does not just return the values set in `fit`, but determines the output feature names from given input feature names if possible.    

__Examples__  
- Target-like encoders encode the features one to one and hence return the same columns as output as the input.  
- Hashing encoding knows how many columns it will produce (and their names) based on `n_components`  
- For BaseN and OneHot encoding the column count depends on the number of distinct values. Hence a not fitted error must be raised.

I'd like to discuss this idea a little with the community. Please add some Pros and Cons of this approach. I'm sure I'm missing something

## Pros
- streamlined behaviour with sklearn. This might lead to better testing and less issues in some corner cases (pipelines etc)
- allows you to get output column names without having to fit an encoder (for some encoders)

## Cons
- quite some work: all encoders need to be changed
- most encoders are 1-1 anyway or would raise a not fitted error.

RNarayan73:
I agree that aligning with the updated sklearn version 1.2 would be very helpful. 

In goes without saying that when using pipelines with multiple transformers, it is useful to have a df at the output end with features traceable to the input features.

The main benefit occurs downstream when the transformed dfs are passed on to predictors. For example, I have a situation currently wherein both XGBT and LGBM can take categorical features directly as an input without having to encode them. They rely on the categorical dtype of the column in the df to identify such features. Without a df output from the transformers in the pipeline with properly labelled columns, it is unnecessarily complicated if not virtually impossible to change the dtype to categorical within the pipeline before it goes on to the predictor.

I was very excited by this new feature in sklearn 1.2 hoping it would deal with it, but not realising that 3rd party transformers also have to play ball and support the api.

I therefore look forward to this being implemented soon!

Narayan

JaimeArboleda:
Yes, I agree with all comments. In fact I am playing with the TargetEncoder and I found that it causes problems when integrated in a Pipeline or in a ColumnTransformer of sklearn, because get_feature_names_out will be called with a parameter and you get this kind of error: 

```
BaseEncoder.get_feature_names_out() takes 1 positional argument but 2 were given
```

JaimeArboleda:
[...]

PR #398, opened 2023-02-23
(WIP) Partial fix for getting feature names out

JaimeArboleda:
I think this is a partial fix for this opened issue:

https://github.com/scikit-learn-contrib/category_encoders/issues/395

It remains to check the behaviour of other estimators that are not ONE_TO_ONE. 

Please, let me know if you like the work in progress and I will try to continue.

### DIFF
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -30,17 +30,6 @@ import sklearn
 import warnings
 from textwrap import dedent
 
-if sklearn.__version__ < '1.2.0':
-    warnings.warn(
-        message=dedent(f"""\
-        You are using version {sklearn.__version__} of scikit-learn.
-        You should consider upgrading the library to at least version "1.2.0".
-        Starting from this version, get_feature_names_out() will be fully supported and will perform \
-        consistently for all estimators even, if they are part of a composed estimator (by \
-        using Pipeline, ColumnTransformer or any other complex object). 
-        """),
-        category=ImportWarning
-    )
 
 __version__ = '2.6.0'
 
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -26,6 +26,29 @@ from category_encoders.cat_boost import CatBoostEncoder
 from category_encoders.rankhot import RankHotEncoder
 from category_encoders.glmm import GLMMEncoder
 from category_encoders.quantile_encoder import QuantileEncoder, SummaryEncoder
+import sklearn
+import warnings
+from textwrap import dedent
+
+if sklearn.__version__ < '1.2.0':
+    warnings.warn(
+        message=dedent(f"""\
+        You are using version {sklearn.__version__} of scikit-learn.
+        You should consider upgrading the library to at least version "1.2.0".
+        Starting from this version, get_feature_names_out() will be fully supported and will perform \
+        consistently for all estimators even, if they are part of a composed estimator (by \
+        using Pipeline, ColumnTransformer or any other complex object). 
+        """),
+        category=ImportWarning
+    )
+else:
+    if sklearn.get_config().get("transform_output", "default") == "default":
+        sklearn.set_config(transform_output="pandas")
+        warnings.warn(
+            message=dedent(f"""scikit-learn output configuration has been set to pandas \
+            to ensure full compatibility of category_encoders with scikit-learn."""),
+            category=RuntimeWarning
+        )
 
 __version__ = '2.6.0'
 
--- a/category_encoders/__init__.py
+++ b/category_encoders/__init__.py
@@ -41,14 +41,6 @@ if sklearn.__version__ < '1.2.0':
         """),
         category=ImportWarning
     )
-else:
-    if sklearn.get_config().get("transform_output", "default") == "default":
-        sklearn.set_config(transform_output="pandas")
-        warnings.warn(
-            message=dedent(f"""scikit-learn output configuration has been set to pandas \
-            to ensure full compatibility of category_encoders with scikit-learn."""),
-            category=RuntimeWarning
-        )
 
 __version__ = '2.6.0'
 
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -358,32 +358,34 @@ class BaseEncoder(BaseEstimator):
         else:
             self.cols = convert_cols_to_list(self.cols)
 
-    def get_feature_names(self) -> List[str]:
+    def get_feature_names(self) -> np.ndarray:
         warnings.warn("`get_feature_names` is deprecated in all of sklearn. Use `get_feature_names_out` instead.",
                       category=FutureWarning)
         return self.get_feature_names_out()
 
-    def get_feature_names_out(self) -> List[str]:
+    def get_feature_names_out(self, input_features=None) -> np.ndarray:
         """
         Returns the names of all transformed / added columns.
 
-        Note that in sklearn the get_feature_names_out function takes the feature_names_in as an argument
-        and determines the output feature names using the input. A fit is usually not necessary and if so a
-        NotFittedError is raised.
-        We just require a fit all the time and return the fitted output columns.
+        It requires a fitted estimator, and it returns the fitted output columns.
 
         Returns
         -------
-        feature_names: list
-            A list with all feature names transformed or added.
+        feature_names: np.ndarray
+            A numpy ndarray with all feature names transformed or added.
             Note: potentially dropped features (because the feature is constant/invariant) are not included!
 
         """
+        if self.encoding_relation != EncodingRelation.ONE_TO_ONE:
+            raise NotImplementedError("This method must be overriden by the Encoder")
         out_feats = getattr(self, "feature_names_out_", None)
         if not isinstance(out_feats, list):
             raise NotFittedError("Estimator has to be fitted to return feature names.")
         else:
-            return out_feats
+            if input_features is not None:
+                return input_features
+            else:
+                return np.array(out_feats)
 
     def get_feature_names_in(self) -> List[str]:
         """
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -260,6 +260,12 @@ class BaseEncoder(BaseEstimator):
         kwargs: dict.
             additional encoder specific parameters like regularisation.
         """
+        if sklearn.__version__ >= '1.2.0':
+            if return_df == False:
+                warnings.warn("This argument will be deprecated (as in newer versions the return value will always be "
+                              "a pandas dataframe)",
+                              category=DeprecationWarning)
+                return_df = True
         self.return_df = return_df
         self.drop_invariant = drop_invariant
         self.invariant_cols = []
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -260,12 +260,6 @@ class BaseEncoder(BaseEstimator):
         kwargs: dict.
             additional encoder specific parameters like regularisation.
         """
-        if sklearn.__version__ >= '1.2.0':
-            if return_df == False:
-                warnings.warn("This argument will be deprecated (as in newer versions the return value will always be "
-                              "a pandas dataframe)",
-                              category=DeprecationWarning)
-                return_df = True
         self.return_df = return_df
         self.drop_invariant = drop_invariant
         self.invariant_cols = []
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -358,7 +358,7 @@ class BaseEncoder(BaseEstimator):
         else:
             self.cols = convert_cols_to_list(self.cols)
 
-    def get_feature_names(self) -> np.ndarray:
+    def get_feature_names(self) -> List[str]:
         warnings.warn("`get_feature_names` is deprecated in all of sklearn. Use `get_feature_names_out` instead.",
                       category=FutureWarning)
         return self.get_feature_names_out()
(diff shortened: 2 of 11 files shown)


category_encoders/utils.py
```diff
@@ -363,7 +363,7 @@ def get_feature_names(self) -> List[str]:
                       category=FutureWarning)
         return self.get_feature_names_out()
 
-    def get_feature_names_out(self) -> List[str]:
+    def get_feature_names_out(self, input_features=None) -> np.ndarray:
```
PaulWestenthanner, line 366: so we're not actually doing anything with the `input_features` parameter? I thought the point of including it would be to explicitly tell every encoder how to calculate output features from the given input features (this is what the 1-1 mixin does, right?)  
For some encoder it give a not-fitted error, but other don't need fitting.  
I feel this is taking some shortcut that sklearn does not take

tests/test_feature_names.py
```diff
@@ -0,0 +1,100 @@
...
+X = X[cat_columns + num_columns]
+np_y = np.random.randn(X.shape[0]) > 0.5
+y = pd.DataFrame(np_y)
+
+class TestEncodersFeaturesOut(TestCase):
+
+    def test_feature_names_out(self):
+        for encoder_name in encoders.__all__:
+            if sklearn.__version__ < "1.2.0":
+                continue
```
PaulWestenthanner, line 28: will the test fail with sklean < 1.2?  
if so I'd force sklearn >= 1.2.0 in the requirement.txt and setup.py

tests/test_rankhot.py
```diff
@@ -45,11 +44,11 @@ def test_extraValue(self):
...
         train_out = enc.fit_transform(train)
-        expected_mapping = pd.DataFrame([[1, 0],[1, 1],], columns=["city_1", "city_2"], index=[1,2])
-        expected_out_train = pd.DataFrame([[1, 0],[1, 1],[1, 0],[1, 1],], columns=["city_1", "city_2"])
-        expected_out_test = pd.DataFrame([[1, 0],[0, 0],], columns=["city_1", "city_2"])
+        expected_mapping = pd.DataFrame([[1, 0], [1, 1], ], columns=["city_1", "city_2"], index=[1, 2])
+        expected_out_train = pd.DataFrame([[1, 0], [1, 1], [1, 0], [1, 1], ], columns=["city_1", "city_2"])
+        expected_out_test = pd.DataFrame([[1, 0], [0, 0], ], columns=["city_1", "city_2"])
         pd.testing.assert_frame_equal(train_out, expected_out_train)
-        pd.testing.assert_frame_equal(enc.mapping[0]["mapping"], expected_mapping)
+        pd.testing.assert_frame_equal(enc.mapping[0]["mapping"], expected_mapping, check_dtype=False)
```
PaulWestenthanner, line 51: why no longer check the dtype? the test hasn't changed except for spacing, has it? why relax it?

tests/test_feature_names.py
```diff
@@ -0,0 +1,100 @@
...
+                    ]
+                )
+                preprocessor = ColumnTransformer(
+                    [
+                        ("categorical_prep_start", categorical_preprocessor_start, ["categorical", "na_categorical"]),
+                        ("categorical_prep_middle", categorical_preprocessor_middle, ["categorical", "na_categorical"]),
+                        ("numerical_prep", numerical_preprocessor, ["float"])
+                    ]
+                )
+                X_tt = preprocessor.fit_transform(X, y)
```
PaulWestenthanner, line 58: this will test both: compatibility with `Pipeline` and `ColumnTransformer`, right?

category_encoders/utils.py
JaimeArboleda: Well, let me explain (for simplicity, every time I say a `Pipeline` it could be that or a `ColumnTransformer` or a `FeatureUnion`...): 
* We need to have the parameter `input_features` because otherwise, when the Encoder is part of a `Pipeline`, there will be an `Exception: BaseEncoder.get_feature_names_out() takes 1 positional argument but 2 were given`, because the method will be called with the parameter `input_features`. 
* Now, what can we do with this parameter? There were two options: modifying all encoders to take into account how to generate the output features given the input_features, or just using the `feature_names_out_`, which is available and computed. Given that this library internally work with dataframes, I thought that the second option was a much better idea. There is only one caveat, which happens again when the encoder is part of a `Pipeline`: if the previous transformer outputted a numpy array, then the column names won't be good (they will be like col_0_A, col_0_B...). That's why I think we should inform that, to ensure full compatibility and 100% correctness of feature names, `sklearn` should be used with the `set_output="pandas"` option. In this case, everything works smoothly, as the tests prove. 
* Another way to put it, in `sklearn` it makes sense to have a function with this signature, because an estimator can be fitted with a numpy array and then it should be able to "compute" the feature names of the output given a list of input_features. But for `category_encoders` this use case does not have too much sense, as the input is expected to be a pandas dataframe. And this assumption is spread everywhere: even other parameters work with column names, like the `mapping` or the `cols`; you won't see parameters expecting column names in `sklearn` because they expect that the input can be a numpy array. 
* So, yes, I agree the behavior is different than in `sklearn`, but I think it makes sense as long as the library is based on this "pandas assumption". I think a big redesign will be a good idea but as we discussed in the issue it will take a lot of effort.

tests/test_feature_names.py
JaimeArboleda: Yes, some tests will fail, because the columns obtained in `get_feature_names_out` won't match the names in the output. But I preferred to put this condition just in case some developer want to run the tests and does not have the latest version of `sklearn`. At least they won't crash.

tests/test_rankhot.py
JaimeArboleda: I checked if all past tests were still running OK, and I found this particular test failed due to some `dtype` issue. What happened was that one column was of `int32` and the other of `int64` or something like that. But the content was the same. So I added this relaxation so that the test passed. Maybe it depends on some pandas version or whatever? I don't know...

tests/test_feature_names.py
JaimeArboleda: Yes, compatibility with both is ensured. And not only feature names: in fact, that nothing crashes...

PaulWestenthanner: nice

tests/test_rankhot.py
PaulWestenthanner: since you haven't really touched this, might just be on your machine, not on the pipeline. But I'm fine with the relaxation

category_encoders/utils.py
PaulWestenthanner: I agree that the proposed change is better than what we have at the moment since it fixes a bug.  
I've got one more question: If we change all encoders to properly work with the input features, will we get compatibility even without setting `set_output=pandas`? Even if we still do all the internal stuff in pandas? If so I think it would be worth it because we'd get compatibility even for sklearn versions < 1.2. This is in fact what I intended when I first raised the issue. In one of the comment I outlined what the changes for each encoder / family of encoders would be. This is far from a complete redesign of the library and also won't get rid of all dataframes. Does this make sense?  
I somehow really don't like introducing dependency versions in the code since this will make it a maintenance hell in the long run

JaimeArboleda: Well, my gut feeling is that it's not worth it: it will still be a very big effort and it will only fix one subtle bug (which doesn't make anything crash) and there will still be some inconsistencies. Let me try to explain... 

* **Current Situation**: If the `category_encoder` is inside a composed `sklearn` transfomer, which delegates the `get_feature_names_out` to every inner estimator, there is an Exception when calling `get_feature_names_out`, because the signature is not valid. The error is `TypeError: get_feature_names_out() takes 1 positional argument but 2 were given`. 
* **Super simple Fix**: By changing the signature but not doing anything with the input parameter, we remove this `Exception` and, making using of the attribute `feature_names_out_`, we get a right answer in most situations. I mean, will be right whenever the `category_encoder` is fitted and transformed alone, or whenever is included in a composed transformer but in the first layer. However, if the `category_encoder` is not in the first layer of a composed transformer, and if `set_output` is the default, then the answer will be wrong, because when fitted the previous transformer of the row will output a numpy array, and the `category_encoder` will take ["0", "1", ...] as the original column names. This is the only *issue* that will remain with this super simple fix (and the solution is as easy as setting `set_output=pandas`). 
* **More thorough fix**:  If we modify every encoder by adding a treatment of the input_features in order to generate the column names (which requieres some work for some of the encoders), the only added value will be removing the wrong answer in the mentioned case (when the encoder is part of a composed transformer and has another `sklearn` transfomer doing something before). My point is that, even in this situation, there will be some inconsistences that are related to the fact of assuming that the input is a pandas dataframe. One of those is the fact that some parameters of the encoders depend on column names. For example, you can specify the mapping for an `OrdinalEncoder` using this: 
    ```python
    OrdinalEncoder(mapping=[{'col': 'col_a', 'mapping': {'a': 2, 'b': 1, 'c': 3}}])
    ``` 
    This code will crash in the same situation as before: if the `OrdinalEncoder` is part o a pipeline and there is another `sklearn` transformer before, the columns received during `fit` won't correspond to the columns specified in the `mapping` parameter (unless `set_output=pandas`). You can take a look at this issue [here](https://colab.research.google.com/drive/1lJJvXyAOoG9MBtNiH2qSTrqbzZtEPTeB#scrollTo=nCJtM_1keNTD&line=2&uniqifier=1). That's why I was saying that if we wish to be fully aligned with `sklearn`, there will be more things to do (in particular, for example, no column names should be used in any parameter of any encoder). So I feel more inclined to say "all or nothing": if we want full compatibility with `sklearn`, that's fine, but it will be a breaking change a lots of work, and otherwise, I think the *super simple fix* is good enough. Well, that's my opinion...

JaimeArboleda: Sorry that I did not answer you question properly: regarding the versions, I would not specify the dependency of 1.2 in the requirements or the setup, because with the simple fix there the remaining bugs are not, in my opinion, a big issue. 

If you want, I could refactor the tests so that I don't require the version for running all tests, but only for some of the asserts (the one that will fail because the column names won't match what `get_feature_names_out` outputs). This way it will be more clear what does not work if not setting the output as pandas.

PaulWestenthanner: > I've got one more question: If we change all encoders to properly work with the input features, will we get compatibility even without setting `set_output=pandas`?

Ok I see, this is not true since the columns to encode are always referenced by name in the `cols` parameter.  
Then I agree with you that changing `get_feature_names_out` to actually do something with the `input_features` is pretty much useless since it won't fix the issue.

PaulWestenthanner: I pretty much agree with you that we should go for the super simple fix. Let me think a little about the version stuff, maybe there is another solution, but I think we're very close to getting this merged

tests/test_feature_names.py
PaulWestenthanner: that's fine

category_encoders/__init__.py
```diff
@@ -26,6 +26,21 @@
 from category_encoders.rankhot import RankHotEncoder
 from category_encoders.glmm import GLMMEncoder
 from category_encoders.quantile_encoder import QuantileEncoder, SummaryEncoder
+import sklearn
+import warnings
+from textwrap import dedent
+
+if sklearn.__version__ < '1.2.0':
```
PaulWestenthanner, line 33: I don't really like the input warning here. I know of some users who use this library in their project and they try to suppress the warning for their end-users.  
Also, at the moment it also does not work anyway. 
I'd be an favor of not issuing a warning here but have something on the `index.rst` (c.f. below comment)

docs/source/index.rst
```diff
@@ -14,7 +14,16 @@ transformers in this library all share a few useful properties:
  * Can explicitly configure which columns in the data are encoded by name or index, or infer non-numeric columns regardless of input type
  * Can drop any columns with very low variance based on training set optionally
  * Portability: train a transformer on data, pickle it, reuse it later and get the same thing out.
- * Full compatibility with sklearn pipelines, input an array-like dataset like any other transformer
+ * Full compatibility with sklearn pipelines, input an array-like dataset like any other transformer (\*)
+
+(\*) For full compatibility with Pipelines and ColumnTransformers, and consistent behaviour of `get_feature_names_out`, it's recommended to upgrade `sklearn` to a version at least '1.2.0' and to set output as pandas:
```
PaulWestenthanner, line 19: could you please introduce a section `known issues` here after usage and before contents, that states something like:

"""
CategoryEncoders internally works with pandas `DataFrames` as apposed to sklearn which works with numpy arrays. This can cause problems in sklearn versions prior to `1.2.0`. In order to ensure full compatibility with sklearn set sklearn to also output `DataFrames`. This can be done by
```python
sklearn.set_config(transform_output="pandas")
```
for a whole project or just for a single pipeline using
```python
Pipeline(
    steps=[
        ("preprocessor", SomePreprocessor().set_output("pandas"),
        ("encoder", SomeEncoder()),
    ]
)
```
If you experience another bug feel free to report it on github https://github.com/scikit-learn-contrib/category_encoders/issues

"""
I think this makes it sufficiently clear and is better than the status quo. Do you agree?

category_encoders/utils.py
PaulWestenthanner: Hi, I've added some thoughts on the version. I'll resolve the discussion here. Please have a look at the other comment I've made and if you're fine with it we're ready to merge!

category_encoders/__init__.py
JaimeArboleda: I fully agree, in fact warnings are annoying. Let's remove this. Do you want me to add a new commit or you prefer to do it in the merge process?

docs/source/index.rst
JaimeArboleda: I think it's very good, yes. Thanks! I don't see any need for more. Let me add this piece and remove the other one so that you can merge :)

category_encoders/utils.py
JaimeArboleda: Hi! Ok, I have added both suggestions!

PaulWestenthanner:
Perfect. Thanks!

PR #409, opened 2023-05-14
Fix issue #407

### LINKED ISSUE #407: Quadratic time intersection on Pandas categories
## Expected Behavior

Using Pandas CategoryDType columns with OrdinalEncoder do not incur a performance penalty.

## Actual Behavior

Pandas' internal categories are intersected with your computed categories in quadratic time, here

https://github.com/scikit-learn-contrib/category_encoders/blob/45110027b263363b53da8564b912e7ca6d5546e8/category_encoders/ordinal.py#L230-L232

## Steps to Reproduce the Problem

  1. Create a Series with a large number of categories, e.g.
  ```python
  categories = [f"Cat{i}" for i in range(10000)]
  series = pd.Series(
      categories,
      pd.CategoricalDtype(categories=categories, ordered=True),
  )
  ```
  1. Apply ordinal encoder to the series

## Specifications

  - Version: 2.5.0
  - Platform:
  - Subsystem:
  
  
 ## Proposed fix
 
 This would be a very simple one line change:
 
 ```python
 categories = list(set(categories).intersection(set(X[col].dtype.categories)))
 ```
 
I'd be happy to get a PR together if this looks okay to you

willsthompson:
Fixes #407

## Proposed Changes

  - Replaces list comprehension intersection with set intersection

### DIFF
--- a/category_encoders/ordinal.py
+++ b/category_encoders/ordinal.py
@@ -229,7 +229,7 @@ class OrdinalEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
                 if util.is_category(X[col].dtype):
                     # Avoid using pandas category dtype meta-data if possible, see #235, #238.
                     if X[col].dtype.ordered:
-                        categories = [c for c in X[col].dtype.categories if c in categories]
+                        categories = list(set(categories).intersection(set(X[col].dtype.categories)))
                     if X[col].isna().any():
                         categories += [np.nan]
 


PaulWestenthanner:
unfortunately the tests are failing, I think this is because your approach might change the order of the categories.
I've followed your idea and just implemented a very basic solution
https://github.com/scikit-learn-contrib/category_encoders/blob/37fcf54613b0a23d52021862d8861600b51dc222/category_encoders/ordinal.py#L232-L233

This should keep the order and since sets are O(1) access time also solve the problem, although it might not be as elegant as yours. If you're happy with it, you can close the issue and PR. I hope I didn't make any mistake here
Thanks for your effort!

PR #410, opened 2023-05-16
replace pkg_resources with importlib.resources

bollwyvl:
Fixes increasingly loud warnings about the deprecation of `pkg_resources`, as seen on https://github.com/conda-forge/category_encoders-feedstock/pull/23

## Proposed Changes

  - replace use of `pkg_resources` with `importlib.metadata`
    - or `importlib_metdata` on python < 3.9
    - also removes an undeclared `setuptools` dependency, which is not _technically_ stdlib

### DIFF
--- a/category_encoders/datasets/_base.py
+++ b/category_encoders/datasets/_base.py
@@ -2,7 +2,11 @@
 Base IO code for datasets
 """
 
-import pkg_resources
+try:
+    from importlib.resources import files, as_file
+except ImportError:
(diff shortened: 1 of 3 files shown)


PaulWestenthanner:
Looks good to me! Thanks for updating this!

bollwyvl:
Thank you! There's a lot of FutureWarnings, too, but they are a little
easier to handle. Maybe get to it another day...

Issue #418, opened 2023-09-03
Support Pandas 2.1


celestinoxp:
WARNING:C:\Users\username\anaconda3\lib\site-packages\category_encoders\utils.py:50: FutureWarning: is_categorical_dtype is deprecated and will be removed in a future version. Use isinstance(dtype, CategoricalDtype) instead
  return pd.api.types.is_categorical_dtype(dtype)

Windows x64
Python 3.10
scikit-learn 1.3
pandas 2.1

PaulWestenthanner:
fixed by #419

PR #419, opened 2023-09-05
Fix is_categorical_dtype deprecation

felunka:
Fixes deprecation warning for pandas "is_categorical_dtype":
category_encoders/utils.py:50: FutureWarning: is_categorical_dtype is deprecated and will be removed in a future version. Use isinstance(dtype, CategoricalDtype) instead

Check:
https://github.com/pandas-dev/pandas/pull/33385

Replacing instances with the recommended alternative to use "isinstance"

### DIFF
--- a/category_encoders/utils.py
+++ b/category_encoders/utils.py
@@ -6,6 +6,7 @@ import warnings
 import pandas as pd
 import numpy as np
 import sklearn.base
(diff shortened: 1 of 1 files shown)


felunka:
Fixes #418

PaulWestenthanner:
Thanks for updating!

PR #425, opened 2023-09-25
Remove uses of inplace

### LINKED ISSUE #422: Pandas copy-on-write doesn't work properly
Here is a basic error message you will get when running (probably) any of the encoders
after setting `pd.options.mode.copy_on_write = True`.
```
ChainedAssignmentError: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
When using the Copy-on-Write mode, such inplace method never works to update the original DataFrame or Series, because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' instead, to perform the operation inplace on the original object.
```

A quick way to fix this automatically would be to run `ruff` with the [PD rules](https://docs.astral.sh/ruff/rules/#pandas-vet-pd) enabled, in particular `PD002` which removes all uses of `inplace`.

s-banach:
Fixes #422 

## Proposed Changes
Remove all 7 uses of `inplace`.

### DIFF
--- a/category_encoders/base_contrast_encoder.py
+++ b/category_encoders/base_contrast_encoder.py
@@ -132,7 +132,7 @@ class BaseContrastEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
 
             # reindex actually applies the mapping
             base_df = mod.reindex(X[col])
-            base_df.set_index(X.index, inplace=True)
+            base_df = base_df.set_index(X.index)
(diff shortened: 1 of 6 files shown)


s-banach:
I will make another PR on top of this one with a few other pandas lints.

PR #426, opened 2023-09-25
Pandas lints

s-banach:
(On top of previous commit regarding `inplace`.)

Most of these changes are simply replacing `.values` with either `.array` or `.to_numpy()`.
The reason `.values` is discouraged, is because it behaves somewhat inconsistently.
- I believe `DataFrame.values` always results a numpy array.
- `Series.values` may return other types of pandas array, such as `pyarrow` arrays.

I've tried to be more explicit, and use `.to_numpy()` whenever the object really needs to be numpy type.

P.S. You can probably ignore these changes if you don't feel comfortable with them, I believe they mostly have no functional impact. The only real useful change is adding `axis=0` to some instances of `np.var()` in the tests. They're going to change the default to `axis=None` in some future version of numpy, which make break those tests.

### DIFF
--- a/category_encoders/base_contrast_encoder.py
+++ b/category_encoders/base_contrast_encoder.py
@@ -132,7 +132,7 @@ class BaseContrastEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
 
             # reindex actually applies the mapping
             base_df = mod.reindex(X[col])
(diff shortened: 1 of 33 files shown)


PaulWestenthanner:
I like this PR! thanks a lot. The library is heavily reliant on pandas and hence it makes sense to follow best practices there!

PR #428, opened 2023-10-08
Optimise `HashingEncoder` for both large and small dataframes

bkhant1:
I used the HashingEncoder recently and found weird that any call to `fit` or `transform`, even for a dataframe with only 10s of rows and a couple of columns took at least 2s... 

I also had quite a large amount of data to encode, and that took a long time. 

That got me started on improving the performance of HashingEncoder, and here's the result! There are quite a few changes in there, each individual change should be in it's own commit, and here's a summary of the performance gain on my machine (macOS Monteray, i7 2.3ghz). 

|  | Baseline | Numpy arrays instead of apply | Shared memory instead of queue | Fork instead of spawn | Faster hashlib usage |
| --- | --- | --- | --- | --- | --- |
| n_rows=30 n_features=3 n_components=10 n_process=4 | 3.55 s ± 150 ms per loop (mean ± std. dev. of ... | 3.62 s ± 140 ms per loop (mean ± std. dev. of ... | 2.2 s ± 41.6 ms per loop (mean ± std. dev. of ... | 56.6 ms ± 2.91 ms per loop (mean ± std. dev. o... | 47.3 ms ± 516 µs per loop (mean ± std. dev. of... |
| n_rows=30 n_features=3 n_components=10 n_process=1 | 1.24 s ± 52.6 ms per loop (mean ± std. dev. of... | 1.42 s ± 170 ms per loop (mean ± std. dev. of ... | 1.74 ms ± 32.2 µs per loop (mean ± std. dev. o... | 2.08 ms ± 91.7 µs per loop (mean ± std. dev. o... | 1.86 ms ± 173 µs per loop (mean ± std. dev. of... |
| n_rows=30 n_features=3 n_components=100 n_process=1 | 1.22 s ± 51.5 ms per loop (mean ± std. dev. of... | 1.33 s ± 60.7 ms per loop (mean ± std. dev. of... | 1.73 ms ± 29.7 µs per loop (mean ± std. dev. o... | 2.01 ms ± 148 µs per loop (mean ± std. dev. of... | 2.01 ms ± 225 µs per loop (mean ± std. dev. of... |
| n_rows=10000 n_features=10 n_components=10 n_process=4 | 5.45 s ± 85.8 ms per loop (mean ± std. dev. of... | 5.36 s ± 57.5 ms per loop (mean ± std. dev. of... | 2.23 s ± 39.6 ms per loop (mean ± std. dev. of... | 120 ms ± 3.02 ms per loop (mean ± std. dev. of... | 96.4 ms ± 2.33 ms per loop (mean ± std. dev. o... |
| n_rows=10000 n_features=10 n_components=10 n_process=1 | 1.61 s ± 30.1 ms per loop (mean ± std. dev. of... | 1.45 s ± 27.2 ms per loop (mean ± std. dev. of... | 227 ms ± 6.03 ms per loop (mean ± std. dev. of... | 236 ms ± 3.06 ms per loop (mean ± std. dev. of... | 170 ms ± 1.35 ms per loop (mean ± std. dev. of... |
| n_rows=100000 n_features=10 n_components=10 n_process=4 | 5.99 s ± 215 ms per loop (mean ± std. dev. of ... | 5.71 s ± 148 ms per loop (mean ± std. dev. of ... | 4.8 s ± 25.4 ms per loop (mean ± std. dev. of ... | 836 ms ± 42.3 ms per loop (mean ± std. dev. of... | 622 ms ± 33.2 ms per loop (mean ± std. dev. of... |
| n_rows=100000 n_features=10 n_components=10 n_process=1 | 5.38 s ± 53 ms per loop (mean ± std. dev. of 7... | 3.73 s ± 56.5 ms per loop (mean ± std. dev. of... | 2.25 s ± 57.4 ms per loop (mean ± std. dev. of... | 3.76 s ± 1.61 s per loop (mean ± std. dev. of ... | 1.68 s ± 19.9 ms per loop (mean ± std. dev. of... |
| n_rows=1000000 n_features=50 n_components=10 n_process=4 | 50.8 s ± 1.17 s per loop (mean ± std. dev. of ... | 56.4 s ± 2.11 s per loop (mean ± std. dev. of ... | 37.1 s ± 576 ms per loop (mean ± std. dev. of ... | 36.9 s ± 2.19 s per loop (mean ± std. dev. of ... | 26.6 s ± 1.8 s per loop (mean ± std. dev. of 7... |
| n_rows=1000000 n_features=50 n_components=10 n_process=1 | 2min 22s ± 2.05 s per loop (mean ± std. dev. o... | 2min 19s ± 3.08 s per loop (mean ± std. dev. o... | 1min 47s ± 1.15 s per loop (mean ± std. dev. o... | 2min 10s ± 18.4 s per loop (mean ± std. dev. o... | 1min 21s ± 1.67 s per loop (mean ± std. dev. o... |

The notebook that produced that table can be found [here](https://gist.github.com/bkhant1/ae2b813817d53b19a81f6774234fcfe3) 

## Proposed Changes

The changes are listed by commit.

### [Add a simple non-regression HashEncoder test](https://github.com/scikit-learn-contrib/category_encoders/commit/0afe06586c71388b8fd4034d196de8a7df4ad56c) 

To make sure I am not breaking it.

### [In HashingEncoder process the df as a numpy array instead of using apply](https://github.com/scikit-learn-contrib/category_encoders/commit/de124410f29778487a2910c8dd7f15ed15785705)

It has no direct impact on performance, however it allows accessing the memory layout of the dataframe directly. That allows using shared memory to communicate between processes instead of a data queue, which does improve performance. 

### [In HashEncoder use shared memory instead of queue for multiproccessing](https://github.com/scikit-learn-contrib/category_encoders/commit/5235a6b85e787b3a384c0d43f314c0e3146d3daf)

It is faster to write directly in memory that to have to data transit through a queue.

The multiprocessing method is similar to what it was with queues: the dataframe is split into chunks, and each process applies the hashing trick to its chunk of the dataframe. Instead of writting the result to a queue, it writes it directly in a shared memory segment, that is also the underlying memory of a numpy array that is used to build the output dataframe.

### [Allow forking processes instead of spwaning them and make it default](https://github.com/scikit-learn-contrib/category_encoders/commit/12f8f242959314ed770750902c1e5ab8ca81263e)

This makes the HashEncoder transform method a lot faster on small datasets.

The spawn process creation method creates a new python interpreter from scratch, and re-import all required module. In a minimal case (pandas and category_encoders.hashing only are imported) this adds a ~2s overhead to any call to transform.

Fork creates a copy of the current process, and that's it. It is unsafe to use with threads, locks, file descriptors, ... but in that case the only thing the forked process will do is process some data and write it to ITS OWN segment of a shared memory. It is a lot faster as pandas doesn't have to be re-imported (around 20ms?)

It might take up more memory as more than the necessary variables (the largest one by far being the HashEncoder instance, which include the user dataframe) will be copied. Add the option to use spawn instead of fork to potentially save some memory.

### [Remove python 2 check code and faster use of hashlib](https://github.com/scikit-learn-contrib/category_encoders/commit/d2d535b4b8b2c54adcb9b13a6b06b5fc8c55286c)

Python 2 is not supported on master, the check isn't useful.

Create int indexes from hashlib bytes digest instead of hex digest as it's faster.

Call the md5 hashlib constructor directly instead of new('md5'), which is also faster.

### DIFF
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -8,6 +8,7 @@ import pandas as pd
 import numpy as np
 import math
 import platform
+from concurrent.futures import ProcessPoolExecutor
 
 __author__ = 'willmcginnis', 'LiuShulun'
 
@@ -108,7 +109,6 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
     """
     prefit_ordinal = False
     encoding_relation = util.EncodingRelation.ONE_TO_M
-    default_int_np_array = np.array(np.zeros((2,2), dtype='int'))
 
     def __init__(self, max_process=0, max_sample=0, verbose=0, n_components=8, cols=None, drop_invariant=False,
                  return_df=True, hash_method='md5', process_creation_method='fork'):
@@ -179,16 +179,17 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
         return X
 
     @staticmethod
-    def hash_chunk(hash_method, shared_memory_result, np_df, N, shared_memory_offset):
+    def hash_chunk(args):
+        hash_method, np_df, N = args
         # Calling getattr outside the loop saves some time in the loop
         hasher_constructor = getattr(hashlib, hash_method)
         # Same when the call to getattr is implicit
         int_from_bytes = int.from_bytes
+        result = np.zeros((np_df.shape[0], N), dtype='int')
         for i, row in enumerate(np_df):
             for val in row:
                 if val is not None:
                     hasher = hasher_constructor()
-                    hasher.update(bytes(str(val), 'utf-8'))
                     # Computes an integer index from the hasher digest. The endian is 
                     # "big" as the code use to read:
                     # column_index = int(hasher.hexdigest(), 16) % N
@@ -196,45 +197,36 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
                     # even if the system is little endian.
                     # Building the index that way is about 30% faster than using the 
                     # hexdigest.
+                    hasher.update(bytes(str(val), 'utf-8'))
                     column_index = int_from_bytes(hasher.digest(), byteorder='big') % N
-                    row_index = (shared_memory_offset + i)*N
-                    shared_memory_index = row_index + column_index
-                    shared_memory_result[shared_memory_index] += 1
+                    result[i, column_index] += 1
+        return result
 
-    def hashing_trick_with_np_parallel(self, df, N):
+    def hashing_trick_with_np_parallel(self, df, N: int):
         np_df = df.to_numpy()
-        shared_memory_result = multiprocessing.RawArray(HashingEncoder.default_int_np_array.dtype.char, len(df)*N)
-
-        process_list = []
-        chunk_size = int(len(np_df)/self.max_process)
         ctx = multiprocessing.get_context(self.process_creation_method)
-        for i in range(0, self.max_process-1):
-            process = ctx.Process(target=self.hash_chunk,
-                args=(self.hash_method, shared_memory_result, np_df[i*chunk_size:((i+1)*chunk_size)], N, i*chunk_size))
-            process_list.append(process)
 
-        # The last process processes all the rest of the dataframe, because the number of rows might not
-        # be divisible by max_process. 
-        process = ctx.Process(target=self.hash_chunk,
-            args=(self.hash_method, shared_memory_result, np_df[(self.max_process-1)*chunk_size:], N, (self.max_process-1)*chunk_size))
-        process_list.append(process)
+        with ProcessPoolExecutor(max_workers=self.max_process, mp_context=ctx) as executor:
+            result = np.concatenate(list(
+                executor.map(
+                    self.hash_chunk,
+                    zip(
+                        [self.hash_method]*self.max_process,
+                        np.array_split(np_df, self.max_process),
+                        [N]*self.max_process
+                    )
+                )
+            ))
 
-        for process in process_list:
-            process.start()
-        for process in process_list:
-            process.join()
-
-        np_result = np.array(shared_memory_result, 'int')
-
-        return pd.DataFrame(np_result.reshape(len(df), N), index=df.index)
+        return pd.DataFrame(result, index=df.index)
 
     def hashing_trick_with_np_no_parallel(self, df, N):
         np_df = df.to_numpy()
-        np_result = np.zeros((len(df)*N), dtype='int')
 
-        HashingEncoder.hash_chunk(self.hash_method, np_result, np_df, N, 0)
+        result = HashingEncoder.hash_chunk((self.hash_method, np_df, N))
+
+        return pd.DataFrame(result, index=df.index)
         
-        return pd.DataFrame(np_result.reshape(len(df), N), index=df.index)
 
     def hashing_trick(self, X_in, hashing_method='md5', N=2, cols=None, make_copy=False):
         """A basic hashing implementation with configurable dimensionality/precision
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -175,12 +175,9 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
         for i, row in enumerate(np_df):
             for val in row:
                 if val is not None:
-                    hasher = hashlib.new('md5')
-                    if sys.version_info[0] == 2:
-                        hasher.update(str(val))
-                    else:
-                        hasher.update(bytes(str(val), 'utf-8'))
-                    column_index = int(hasher.hexdigest(), 16) % N
+                    hasher = hashlib.md5()
+                    hasher.update(bytes(str(val), 'utf-8'))
+                    column_index = int.from_bytes(hasher.digest(), byteorder='big') % N
                     row_index = (shm_offset + i)*N
                     shm_index = row_index + column_index
                     shm_result[shm_index] += 1
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -122,7 +122,6 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
             self.max_process = max_process
         self.max_sample = int(max_sample)
         self.process_creation_method = process_creation_method
-        self.auto_sample = max_sample <= 0
         self.data_lines = 0
         self.X = None
 
@@ -171,13 +170,24 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
         return X
 
     @staticmethod
-    def hash_chunk(shm_result, np_df, N, shm_offset):
+    def hash_chunk(hash_method, shm_result, np_df, N, shm_offset):
+        # Calling getattr outside the loop saves some time in the loop
+        hasher_constructor = getattr(hashlib, hash_method)
+        # Same when the call to getattr is implicit
+        int_from_bytes = int.from_bytes
         for i, row in enumerate(np_df):
             for val in row:
                 if val is not None:
-                    hasher = hashlib.md5()
+                    hasher = hasher_constructor()
                     hasher.update(bytes(str(val), 'utf-8'))
-                    column_index = int.from_bytes(hasher.digest(), byteorder='big') % N
+                    # Computes an integer index from the hasher digest. The endian is 
+                    # "big" as the code use to read:
+                    # column_index = int(hasher.hexdigest(), 16) % N
+                    # which is implicitly considering the hexdigest to be big endian,
+                    # even if the system is little endian.
+                    # Building the index that way is about 30% faster than using the 
+                    # hexdigest.
+                    column_index = int_from_bytes(hasher.digest(), byteorder='big') % N
                     row_index = (shm_offset + i)*N
                     shm_index = row_index + column_index
                     shm_result[shm_index] += 1
@@ -186,35 +196,34 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
         np_df = df.to_numpy()
         shm_result = multiprocessing.RawArray(HashingEncoder.default_int_np_array.dtype.char, len(df)*N)
 
-        n_process = []
+        process_list = []
         chunk_size = int(len(np_df)/self.max_process)
         ctx = multiprocessing.get_context(self.process_creation_method)
         for i in range(0, self.max_process-1):
             process = ctx.Process(target=self.hash_chunk,
-                args=(shm_result, np_df[i*chunk_size:((i+1)*chunk_size)], N, i*chunk_size))
-            n_process.append(process)
+                args=(self.hash_method, shm_result, np_df[i*chunk_size:((i+1)*chunk_size)], N, i*chunk_size))
+            process_list.append(process)
 
         # The last process processes all the rest of the dataframe, because the number of rows might not
         # be divisible by max_process. 
         process = ctx.Process(target=self.hash_chunk,
-            args=(shm_result, np_df[(self.max_process-1)*chunk_size:], N, (self.max_process-1)*chunk_size))
-        n_process.append(process)
+            args=(self.hash_method, shm_result, np_df[(self.max_process-1)*chunk_size:], N, (self.max_process-1)*chunk_size))
+        process_list.append(process)
 
-        for process in n_process:
+        for process in process_list:
             process.start()
-        for process in n_process:
+        for process in process_list:
             process.join()
 
         np_result = np.array(shm_result, 'int')
 
         return pd.DataFrame(np_result.reshape(len(df), N), index=df.index)
 
-    @staticmethod
-    def hashing_trick_with_np_no_parallel(df, N):
+    def hashing_trick_with_np_no_parallel(self, df, N):
         np_df = df.to_numpy()
         np_result = np.zeros((len(df)*N), dtype='int')
 
-        HashingEncoder.hash_chunk(np_result, np_df, N, 0)
+        HashingEncoder.hash_chunk(self.hash_method, np_result, np_df, N, 0)
         
         return pd.DataFrame(np_result.reshape(len(df), N), index=df.index)
 
--- a/category_encoders/hashing.py
+++ b/category_encoders/hashing.py
@@ -102,6 +102,7 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
     """
     prefit_ordinal = False
     encoding_relation = util.EncodingRelation.ONE_TO_M
+    default_int_np_array = np.array(np.zeros((2,2), dtype='int'))
 
     def __init__(self, max_process=0, max_sample=0, verbose=0, n_components=8, cols=None, drop_invariant=False,
                  return_df=True, hash_method='md5'):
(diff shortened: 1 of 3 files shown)


category_encoders/hashing.py
```diff
@@ -129,87 +132,7 @@ def __init__(self, max_process=0, max_sample=0, verbose=0, n_components=8, cols=
...
-        self.X = X
-
-        self.data_lines = len(self.X)
-
-        data_lock = multiprocessing.Manager().Lock()
-        new_start = multiprocessing.Manager().Value('d', True)
-        done_index = multiprocessing.Manager().Value('d', int(0))
-        hashing_parts = multiprocessing.Manager().Queue()
-
-        if self.auto_sample:
```
PaulWestenthanner, line 179: you don't need the auto sample attribute anymore

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
+                    column_index = int.from_bytes(hasher.digest(), byteorder='big') % N
+                    row_index = (shm_offset + i)*N
+                    shm_index = row_index + column_index
+                    shm_result[shm_index] += 1
+
+    def hashing_trick_with_np_parallel(self, df, N):
+        np_df = df.to_numpy()
+        shm_result = multiprocessing.RawArray(HashingEncoder.default_int_np_array.dtype.char, len(df)*N)
+
+        n_process = []
```
PaulWestenthanner, line 189: `n_process` sounds like an integer. Better call it `process_list` or something more telling. You only copied the name but probably it's time to change it now

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
+        chunk_size = int(len(np_df)/self.max_process)
```
PaulWestenthanner, line 190: this ignores the `max_samples` parameter and might lead to the process crashing in case of too much data / too few CPUs

category_encoders/hashing.py
```diff
@@ -101,9 +102,10 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
     """
     prefit_ordinal = False
     encoding_relation = util.EncodingRelation.ONE_TO_M
+    default_int_np_array = np.array(np.zeros((2,2), dtype='int'))
 
     def __init__(self, max_process=0, max_sample=0, verbose=0, n_components=8, cols=None, drop_invariant=False,
-                 return_df=True, hash_method='md5'):
+                 return_df=True, hash_method='md5', process_creation_method='fork'):
```
PaulWestenthanner, line 108: please add the `process_creation_method` parameter in the documentation and explain the options

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
+    def hash_chunk(shm_result, np_df, N, shm_offset):
+        for i, row in enumerate(np_df):
+            for val in row:
+                if val is not None:
+                    hasher = hashlib.md5()
+                    hasher.update(bytes(str(val), 'utf-8'))
+                    column_index = int.from_bytes(hasher.digest(), byteorder='big') % N
+                    row_index = (shm_offset + i)*N
+                    shm_index = row_index + column_index
+                    shm_result[shm_index] += 1
```
PaulWestenthanner, line 183: how is this the same as the old code? wasn't the old code just doing `shm_result[column_index] += 1`?

category_encoders/hashing.py
```diff
@@ -101,9 +102,10 @@ class HashingEncoder(util.BaseEncoder, util.UnsupervisedTransformerMixin):
     """
     prefit_ordinal = False
     encoding_relation = util.EncodingRelation.ONE_TO_M
+    default_int_np_array = np.array(np.zeros((2,2), dtype='int'))
```
PaulWestenthanner, line 105: why is this 2x2?

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
+                    shm_result[shm_index] += 1
+
+    def hashing_trick_with_np_parallel(self, df, N):
+        np_df = df.to_numpy()
+        shm_result = multiprocessing.RawArray(HashingEncoder.default_int_np_array.dtype.char, len(df)*N)
+
+        n_process = []
+        chunk_size = int(len(np_df)/self.max_process)
+        ctx = multiprocessing.get_context(self.process_creation_method)
+        for i in range(0, self.max_process-1):
```
PaulWestenthanner, line 192: splitting this way is not very elegant if the last chunk has to be treated separately.
using numpy array split could be helpful: https://stackoverflow.com/a/75981560.

Wouldn't it be easier if the `hash_chunk` function would hash a chunk and return an array. Then it wouldn't need the `shm_result` and `shm_offset` parameters (what does shm stand for btw?). Then you'd just concatenate all the chunks in the end?

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
+
+        return pd.DataFrame(np_result.reshape(len(df), N), index=df.index)
 
     @staticmethod
-    def hashing_trick(X_in, hashing_method='md5', N=2, cols=None, make_copy=False):
+    def hashing_trick_with_np_no_parallel(df, N):
+        np_df = df.to_numpy()
+        np_result = np.zeros((len(df)*N), dtype='int')
+
+        HashingEncoder.hash_chunk(np_result, np_df, N, 0)
```
PaulWestenthanner, line 217: do you need to re-assign this to `np_result` or will this be updated in place?

PaulWestenthanner (commented):
Thanks for that contribution! Seems to improve the times a lot. 
I tried to understand everything and left some comments where further clarification is necessary

PaulWestenthanner:
one more question: you've removed all the data-lock stuff for multi-processing, right. Was this not necessary?

category_encoders/hashing.py
bkhant1: Removing it!

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
-        else:
-            return X.to_numpy()
+    @staticmethod
+    def hash_chunk(shm_result, np_df, N, shm_offset):
+        for i, row in enumerate(np_df):
+            for val in row:
+                if val is not None:
+                    hasher = hashlib.md5()
+                    hasher.update(bytes(str(val), 'utf-8'))
+                    column_index = int.from_bytes(hasher.digest(), byteorder='big') % N
```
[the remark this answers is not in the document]
bkhant1: I will add a comment! The `digest` version is about 30% faster than the `hexdigest` version. On my machine:

```python
> import hashlib
> hasher = hashlib.md5()
> hasher.update("abdcde1234")
> %timeit int(hasher.hexdigest(), 16)
659 ns ± 29.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
> %timeit int.from_bytes(hasher.digest(), byteorder='big')
518 ns ± 11.5 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

The byte order is machine dependant only if the data comes directly from the machine. In that case, the `digest` should not be machine dependant. I tried it on
- my system (MacOS Monterey little endian) 
```python
> import sys
> import hashlib
> sys.byteorder
'little'
> hasher = hashlib.md5()
> hasher.update("abc")
> int.from_bytes(hasher.digest(), endian='big')
191415658344158766168031473277922803570
> int(hasher.hexdigest(), 16)
191415658344158766168031473277922803570
```
- a docker system (`s390x/ubuntu` image, big endian):
```python
> import sys
> import hashlib
> sys.byteorder
'big'
> hasher = hashlib.md5()
> hasher.update("abc")
> int.from_bytes(hasher.digest(), endian='big')
191415658344158766168031473277922803570
> int(hasher.hexdigest(), 16)
191415658344158766168031473277922803570
```

The result is the same.

category_encoders/hashing.py
bkhant1: That is actually another advantage of using a shared memory instead of queues. 

In the previous implementation, there were 3 mains things that could take up memory:
1. the input dataframe, stored in `self.X` - it is copied accross all processes wether we fork or spawn because the process target is `self.require_data`. 
2. the output dataframe, which is the concatenation of all `hashing_parts`. There is only one instance of it in the main process. It takes space of exactly the output dataframe when all processes are done computing their chunk of the data.
3. the `data_part` that's temporary to each subprocess, and that's created to be put on the queue after the part is calculated.

In the worst case scenario of the queue implementation, the total memory usage is: `input_size * number_of_processes + output_size + (ouput_size/max_sample)*number_of_processes`

In the new implementation, only two things, because we write directly to the output shared memory instead of having the data transiting through the queue:
1. the input dataframe, stored in `self.X` - it is copied accross all processes wether we fork or spawn because the process has in its args `np_df`.
2. the output dataframe, which is the shared memory and is allocated only accross all processes.

The total memory usage in the shared memory implementation is: `imput_size * number_of_processes + output_size`

That means that the memory usage is always smaller with the shared memory implementation, and does not depend on `max_samples`

category_encoders/hashing.py
bkhant1: It was, but it was in the context of writting to a subset of the output dataframe. In that case we write to the "global" output dataframe, so we need to add the offset specific to the chunk we're processing.

category_encoders/hashing.py
bkhant1: `shm` stands for "shared_memory" - will update the variable name 👍 

> Wouldn't it be easier if the hash_chunk function would hash a chunk and return an array. Then it wouldn't need the shm_result and shm_offset parameters (what does shm stand for btw?). Then you'd just concatenate all the chunks in the end?

That's a very good point, I will try that

category_encoders/hashing.py
```diff
@@ -238,18 +161,64 @@ def _transform_single_cpu(self, X, override_return_df=False):
...
-        if self.return_df or override_return_df:
-            return X
-        else:
-            return X.to_numpy()
+    @staticmethod
+    def hash_chunk(shm_result, np_df, N, shm_offset):
+        for i, row in enumerate(np_df):
+            for val in row:
+                if val is not None:
+                    hasher = hashlib.md5()
```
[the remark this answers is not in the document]
bkhant1: My bad thanks for catching that! I'm replacing it with `hash_ctor = getattr(hashlib, hash_key)` outside the loop and `hash_ctor()` in the loop which is even faster! 

```python
> ctor = getattr(hashlib, 'md5')
> %timeit ctor()
190 ns ± 0.927 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
> %timeit hashlib.md5()
231 ns ± 4.86 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

category_encoders/hashing.py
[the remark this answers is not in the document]
PaulWestenthanner: Ok that makes sense.

Which python/hashlib version are you using btw?
Running `hasher.update("abc")` results in a `TypeError` on my machine since it requires (as per documentation) bytes not strings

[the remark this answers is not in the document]
bkhant1: I copied the wrong line from my shell! I updated my comment above.

category_encoders/hashing.py
bkhant1: Im not sure 😅 I am just using it to get the type of the default int array name in [that line](https://github.com/scikit-learn-contrib/category_encoders/pull/428/files/206de2d8327489c4ff4d2c7c4f566c0fc06210c1#diff-5871b042f65ccab77377b2e9a92ea2c9651cc039b020835b6d77bfcb01ffe475R187) so it could be 1 by 1!

category_encoders/hashing.py
bkhant1: It gets updated in place!

bkhant1:
Hi @PaulWestenthanner ! Thanks for your review!

> one more question: you've removed all the data-lock stuff for multi-processing, right. Was this not necessary?

It was necessary because all processes were writting to the same queue. With shared memory they write to their own section of the shared memory, so no concurrency to manage there.

I think I replied to all your comments/addressed them. 

The one outstanding thing is that you suggested:
> Wouldn't it be easier if the hash_chunk function would hash a chunk and return an array. Then it wouldn't need the shm_result and shm_offset parameters (what does shm stand for btw?). Then you'd just concatenate all the chunks in the end?

It's a little more complex than that because `return` doesn't really work accross processes (hence the use of queues or shared memory), but I remembered that python has a nice API, [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor), (that uses queues behind the scenes) to make it look like it's possible to return something from a process. 

Here's [a PR](https://github.com/bkhant1/category_encoders/compare/all_optis...bkhant1:category_encoders:multiproc_pool?expand=1) in my fork showing what I tried. It could be a little slower but the code is cleaner. I think I should add this commit to this PR but let me know what you think!

|Baseline                                                |Before review                                    |After review                                     |Multiproc pool instead of SHM                    |FIELD5                                           |
|--------------------------------------------------------|-------------------------------------------------|-------------------------------------------------|-------------------------------------------------|-------------------------------------------------|
|n_rows=30 n_features=3 n_components=10 n_process=4      |3.12 s ± 20.2 ms per loop (mean ± std. dev. of...|53.9 ms ± 443 µs per loop (mean ± std. dev. of...|55 ms ± 1.21 ms per loop (mean ± std. dev. of ...|59.8 ms ± 989 µs per loop (mean ± std. dev. of...|
|n_rows=30 n_features=3 n_components=10 n_process=1      |1.1 s ± 17 ms per loop (mean ± std. dev. of 7 ...|1.56 ms ± 45.5 µs per loop (mean ± std. dev. o...|1.5 ms ± 44.8 µs per loop (mean ± std. dev. of...|1.52 ms ± 39.8 µs per loop (mean ± std. dev. o...|
|n_rows=30 n_features=3 n_components=100 n_process=1     |1.1 s ± 14.5 ms per loop (mean ± std. dev. of ...|1.52 ms ± 25.2 µs per loop (mean ± std. dev. o...|1.77 ms ± 317 µs per loop (mean ± std. dev. of...|1.49 ms ± 21.7 µs per loop (mean ± std. dev. o...|
|n_rows=10000 n_features=10 n_components=10 n_process=4  |4.75 s ± 36 ms per loop (mean ± std. dev. of 7...|99.3 ms ± 649 µs per loop (mean ± std. dev. of...|122 ms ± 13 ms per loop (mean ± std. dev. of 7...|113 ms ± 992 µs per loop (mean ± std. dev. of ...|
|n_rows=10000 n_features=10 n_components=10 n_process=1  |1.51 s ± 13.5 ms per loop (mean ± std. dev. of...|156 ms ± 2.67 ms per loop (mean ± std. dev. of...|165 ms ± 17.6 ms per loop (mean ± std. dev. of...|146 ms ± 1.79 ms per loop (mean ± std. dev. of...|
|n_rows=100000 n_features=10 n_components=10 n_process=4 |5.53 s ± 30.6 ms per loop (mean ± std. dev. of...|584 ms ± 6.74 ms per loop (mean ± std. dev. of...|576 ms ± 25.2 ms per loop (mean ± std. dev. of...|572 ms ± 12.4 ms per loop (mean ± std. dev. of...|
|n_rows=100000 n_features=10 n_components=10 n_process=1 |5.22 s ± 55.3 ms per loop (mean ± std. dev. of...|1.55 s ± 11 ms per loop (mean ± std. dev. of 7...|1.48 s ± 12.6 ms per loop (mean ± std. dev. of...|1.46 s ± 32.7 ms per loop (mean ± std. dev. of...|
|n_rows=1000000 n_features=50 n_components=10 n_process=4|53.2 s ± 2.35 s per loop (mean ± std. dev. of ...|23.5 s ± 256 ms per loop (mean ± std. dev. of ...|22.3 s ± 200 ms per loop (mean ± std. dev. of ...|25.6 s ± 1.6 s per loop (mean ± std. dev. of 7...|

PaulWestenthanner:
Thanks for the changes. Looks really cool now. 
I'd prefer the clean `ProcessPoolExecutor` implementation. Then we'll be able to merge!

PaulWestenthanner:
could you maybe also add a high level summary to the changelog in the unreleased section (https://github.com/scikit-learn-contrib/category_encoders/blob/master/CHANGELOG.md)?

bkhant1:
I added the `ProcessPoolExecutor` commit to this branch and updated the changelog!

PaulWestenthanner:
Looks good! Thanks for this improvement! Makes it both more readable and faster

PR #431, opened 2023-11-21
if y is bool convert it into numerical

mausam3407:
Fixes #

## Proposed Changes

We're converting the y to float in _transform function but not in fit function. And for other transforms as well, if we're required to do numerical calculation on y, we first convert it to float. This PR is to maintain the consitency in behaviour
  -
  -
  -


PaulWestenthanner:
nice idea, LGTM & merging

PR #432, opened 2023-12-06
Fixed the DataFrame conversion of y in PolynomialWrapper's fit method

Asiaticum:
Fixes #

## Proposed Changes
Changed from y = pd.DataFrame(y, columns=['target']) to y = pd.DataFrame(y.rename('target')).
This is because if the original y is a pd.Series with a name other than 'target', pd.DataFrame(y, columns=['target']) returns an empty DataFrame.

The root cause of this problem is that in the function of utils.convert_input_vector(y, index) -> pd.Series, when a pd.Series or a single-column pd.DataFrame is passed as an argument, the process is designed to inherit the column name from the column name of the passed pd.Series or pd.DataFrame.


PaulWestenthanner:
Neat! well spotted. I'll merge it if the pipeline succeeds

Issue #445, opened 2024-10-01
Add python 3.11 and 3.12 to build


PaulWestenthanner:
## Expected Behavior

The pipeline should also run python 3.11 and python 3.12

Buedenbender:
+1 would be highly appreciated

PaulWestenthanner:
Just released this in version 2.7.0

PR #447, opened 2024-12-05
chore(ci): added python3.11 and python3.12 tests

marekschneider:
## Proposed Changes

  - Added Python 3.11 and 3.12 into the Pytest-Matrix (https://github.com/scikit-learn-contrib/category_encoders/issues/445)

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
 
     steps:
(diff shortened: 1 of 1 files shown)


Buedenbender:
Highly appreciated and anticipated.
Thank you very much

PaulWestenthanner:
The github action pipeline fails as python 3.7 is no longer supported by them.
I just checked and python 3.8 also reached the end of life, so we should probably remove 3.7 and 3.8

PaulWestenthanner:
seems like we're not compatible with sklearn >= 1.6.

Quick fix would be to specify this in requirements.txt and create an issue (or properly fit it).

I'm currently working on a rather big change that should add type hinting, update all dependencies and re-work the tooling including a pipeline that does type checking and format checking etc.
https://github.com/PaulWestenthanner/categorical-encoding/tree/improve_type_hints

PaulWestenthanner:
I've added support. Closing this now.

PR #461, opened 2025-08-15
Fix TargetEncoder compatibility with scikit-learn ColumnTransformer

### LINKED ISSUE #460: TargetEncoder cannot be cloned by scikit-learn when hierarchy is passed
# Description

scikit-learn expects to be able to clone transformers in many contexts including the `ColumnTransformer`.  It will throw an exception if the parameters do not match the original parameters.  The `TargetEncoder` modifies the `hierarchy` object passed if it is not None, and this causes an exception when the `TargetEncoder` is used in a `ColumnTransformer` and likely other contexts.

From the [scikit-learn docs](https://scikit-learn.org/stable/developers/develop.html#instantiation:~:text=There%20should%20be,in%20fit.):

> There should be no logic, not even input validation, and the parameters should not be changed; which also means ideally they should not be mutable objects such as lists or dictionaries. If they’re mutable, they should be copied before being modified. The corresponding logic should be put where the parameters are used, typically in fit.

# Example

Here's an example:
```python
import pandas as pd
from sklearn.compose import ColumnTransformer

features: list[str] = ["cat_feature", "int_feature"]
target: str = "target"

df: pd.DataFrame = pd.DataFrame({
    "cat_feature": ["aa", "ab", "ac", "ba", "bb", "bc"],
    "int_feature": [10, 11, 9, 8, 12, 10],
    "target": [1, 2, 1, 2, 1, 2]
})

hierarchical_map: dict = {
    "cat_feature": {
        "a": ("aa", "ab", "ac"),
        "b": ("ba", "bb", "bc"),
    },
}

# This works
target_encoder = ce.TargetEncoder(cols=["cat_feature"], min_samples_leaf=2, smoothing=2, hierarchy=hierarchical_map)
target_encoder.fit_transform(df[features], df[target])

# This fails
preprocessor = ColumnTransformer(
    transformers=[
        ("categorical", ce.TargetEncoder(cols=["cat_feature"], min_samples_leaf=2, smoothing=2, hierarchy=hierarchical_map), ["cat_feature"]),
    ],
    remainder='passthrough',
    verbose_feature_names_out=False,
)
preprocessor.fit_transform(df[features], df[target])
```

The failure message:
```
RuntimeError: Cannot clone object TargetEncoder(cols=['cat_feature'],
              hierarchy={'cat_feature': {'aa': ('a',), 'ab': ('a',),
                                         'ac': ('a',), 'ba': ('b',),
                                         'bb': ('b',), 'bc': ('b',)}},
              min_samples_leaf=2, smoothing=2), as the constructor either does not set or modifies parameter hierarchy
```

# Work-around

I can work around this by creating a class that inherits from `TargetEncoder` but moves the logic from the `__init__` to the `fit` method.  This is slightly modified to not alter the original `self.hierarchy`.  Instead, it creates a new object, `inverted_hierarchy` in the `fit` method.  (BTW, I'm not sure what a good name for this should be.)

```python
import category_encoders.utils as util
from category_encoders.ordinal import OrdinalEncoder


class ModifiedTargetEncoder(ce.TargetEncoder):
    def __init__(
        self,
        verbose: int = 0,
        cols: list[str] = None,
        drop_invariant: bool = False,
        return_df: bool = True,
        handle_missing: str = 'value',
        handle_unknown: str = 'value',
        min_samples_leaf: int = 20,
        smoothing: float = 10,
        hierarchy: dict = None,
    ) -> None:
        super(ce.TargetEncoder, self).__init__(
            verbose=verbose,
            cols=cols,
            drop_invariant=drop_invariant,
            return_df=return_df,
            handle_unknown=handle_unknown,
            handle_missing=handle_missing,
        )
        self.min_samples_leaf = min_samples_leaf
        self.smoothing = smoothing
        self.hierarchy = hierarchy
        self.ordinal_encoder = None
        self.mapping = None
        self._mean = None

    def _fit(self, X: util.X_type, y: util.y_type, **kwargs) -> None:
        # BEGIN copied from the original __init__
        # @ToDo create a function to check the hierarchy
        if isinstance(self.hierarchy, (dict, pd.DataFrame)) and self.cols is None:
            raise ValueError('Hierarchy is defined but no columns are named for encoding')
        if isinstance(self.hierarchy, dict):
            inverted_hierarchy = {}
            self.hierarchy_depth = {}
            for switch in self.hierarchy:
                flattened_hierarchy = util.flatten_reverse_dict(self.hierarchy[switch])
                hierarchy_check = self._check_dict_key_tuples(flattened_hierarchy)
                self.hierarchy_depth[switch] = hierarchy_check[1]
                if not hierarchy_check[0]:
                    raise ValueError(
                        'Hierarchy mapping contains different levels for key "' + switch + '"'
                    )
                inverted_hierarchy[switch] = {
                    (k if isinstance(t, tuple) else t): v
                    for t, v in flattened_hierarchy.items()
                    for k in t
                }
        elif isinstance(hierarchy, pd.DataFrame):
            inverted_hierarchy = self.hierarchy
            self.hierarchy_depth = {}
            for col in self.cols:
                HIER_cols = inverted_hierarchy.columns[
                    inverted_hierarchy.columns.str.startswith(f'HIER_{col}')
                ].tolist()
                HIER_levels = [int(i.replace(f'HIER_{col}_', '')) for i in HIER_cols]
                if np.array_equal(sorted(HIER_levels), np.arange(1, max(HIER_levels) + 1)):
                    self.hierarchy_depth[col] = max(HIER_levels)
                else:
                    raise ValueError(f'Hierarchy columns are not complete for column {col}')
        elif self.hierarchy is None:
            inverted_hierarchy = hierarchy
        else:
            raise ValueError('Given hierarchy mapping is neither a dictionary nor a dataframe')

        self.cols_hier = []
        # END copied from the original __init__

        # BEGIN copied from the original _fit
        if isinstance(inverted_hierarchy, dict):
            X_hier = pd.DataFrame()
            for switch in inverted_hierarchy:
                if switch in self.cols:
                    colnames = [
                        f'HIER_{str(switch)}_{str(i + 1)}'
                        for i in range(self.hierarchy_depth[switch])
                    ]
                    df = pd.DataFrame(
                        X[str(switch)].map(inverted_hierarchy[str(switch)]).tolist(),
                        index=X.index,
                        columns=colnames,
                    )
                    X_hier = pd.concat([X_hier, df], axis=1)
        elif isinstance(inverted_hierarchy, pd.DataFrame):
            X_hier = inverted_hierarchy

        if isinstance(inverted_hierarchy, (dict, pd.DataFrame)):
            enc_hier = OrdinalEncoder(
                verbose=self.verbose,
                cols=X_hier.columns,
                handle_unknown='value',
                handle_missing='value',
            )
            enc_hier = enc_hier.fit(X_hier)
            X_hier_ordinal = enc_hier.transform(X_hier)

        self.ordinal_encoder = OrdinalEncoder(
            verbose=self.verbose, cols=self.cols, handle_unknown='value', handle_missing='value'
        )
        self.ordinal_encoder = self.ordinal_encoder.fit(X)
        X_ordinal = self.ordinal_encoder.transform(X)
        if inverted_hierarchy is not None:
            self.mapping = self.fit_target_encoding(
                pd.concat([X_ordinal, X_hier_ordinal], axis=1), y
            )
        else:
            self.mapping = self.fit_target_encoding(X_ordinal, y)
         # END copied from the original _fit

# This works
modified_preprocessor = ColumnTransformer(
    transformers=[
        ("categorical", ModifiedTargetEncoder(cols=["cat_feature"], min_samples_leaf=2, smoothing=2, hierarchy=hierarchical_map), ["cat_feature"]),
    ],
    remainder='passthrough',
    verbose_feature_names_out=False,
).set_output(transform="pandas")
modified_preprocessor.fit_transform(df[features], df[target])
```

I'd be happy to make a PR.  But since I'm new to this project, I wanted to make sure I understood things first.

# Related

This might be related to https://github.com/scikit-learn-contrib/category_encoders/issues/184 which covers problems using `TargetEncoder` with `RandomizedSearch` from `scikit-learn`.  But that did not involve passing a `hierarchy` object, and it was resolved.

dennisobrien:
Fixes #460 

## Proposed Changes

- This fixes a bug where a `TargetEncoder` passed a non-null `hierarchy` object would throw an exception when used in a `scikit-learn` `ColumnTransformer`.
- This moves some logic and validation from the `TargetEncoder` `__init__` method to the `fit` method in order to pass validation tests in the `scikit-learn` `clone` method.
- Added a test for this scenario: `test_hierarchy_with_scikit_learn_column_transformer`.


category_encoders/target_encoder.py
```diff
@@ -344,3 +311,43 @@ def _weighting(self, n: int) -> float:
         # monotonically increasing function of n bounded between 0 and 1
         # sigmoid in this case, using scipy.expit for numerical stability
         return expit((n - self.min_samples_leaf) / self.smoothing)
+
+    def _generate_inverted_hierarchy(self) -> tuple[dict | pd.DataFrame, dict]:
```
PaulWestenthanner: this returns a `tuple[dict | pd.DataFrame | None, dict]` right? If self.hierarchy is None this will also be None.
I appreciate the type hints btw. I'm working on a branch to add typing support to the whole library.

category_encoders/target_encoder.py
```diff
@@ -219,24 +185,25 @@ def _check_dict_key_tuples(dict_to_check: dict[Any, tuple]) -> tuple[bool, int]:
         return min_tuple_size == max_tuple_size, min_tuple_size
 
     def _fit(self, X: util.X_type, y: util.y_type, **kwargs) -> None:
-        if isinstance(self.hierarchy, dict):
+        inverted_hierarchy, self.hierarchy_depth = self._generate_inverted_hierarchy()
```
PaulWestenthanner: is there a reason for setting `self.hierarchy_depth` instead of just keeping it a variable in the scope of the `_fit` function?

dennisobrien: `self.hierarchy_depth` is used by `fit_target_encoding` internally.  That function looks like it's part of the public API, but I don't see it explicitly tested in `tests/test_target_encoder.py` and I'm confused how it would work without doing all the setup work handled in `_fit`.

Changing `self.hierarchy_depth` to a local variable scoped in `_fit` would require it to be recomputed in `fit_target_encoding`.

category_encoders/target_encoder.py
dennisobrien: Good catch.  I made this change.

category_encoders/target_encoder.py
```diff
@@ -344,3 +311,43 @@ def _weighting(self, n: int) -> float:
         # monotonically increasing function of n bounded between 0 and 1
         # sigmoid in this case, using scipy.expit for numerical stability
         return expit((n - self.min_samples_leaf) / self.smoothing)
+
+    def _generate_inverted_hierarchy(self) -> tuple[dict | pd.DataFrame, dict]:
+        # @ToDo create a function to check the hierarchy
```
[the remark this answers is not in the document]
dennisobrien: I removed the TODO comment.

dennisobrien:
> This fix looks great, code is cleaner than before, test is also nice! Great job and sorry again for having you wait so long. Can you please also rebase with the current master. I've updated some sphinx versions in order for the CI pipeline to work, it somehow broke down with the old versions.

Thanks for the feedback.  And no problem about the timeline.  I appreciate your contributions in creating and maintaining this project.

I made the changes you mentioned and followed up in each of the comments.

Let me know if there is anything else you'd like me to change in this.

